PR #31, opened 2018-12-07
build: Name firmware bin and elf file by the name of the firmware

ronald-cron-arm:
Having different file names makes it easier to differentiate
the firmware binaries. It also removes the need for renaming
them when they are placed in a common directory.

The firmware.bin binary file is still generated by the build
system. A subsequent patch will remove its generation giving
time to remove any dependency on it.

Change-Id: Id38c20d667f802f18abf090b0c331b796c2ce23d
Signed-off-by: Ronald Cron <ronald.cron@arm.com>


ronald-cron-arm:
jenkins: test this please

PR #32, opened 2018-12-20
Deepan02 arm n1sdp scp mcp rom fw

deepan02:
patch set for
-> addition of scp/mcp rom firmware.
-> updates to copy the scp/mcp ram fw .


davidcunado-arm:
@deepan02 sorry to be a pain, but please could you split these patches into 2 PRs - one for the rom firmware and another for the misc updates. 

The rationale is that I'd to keep large bits of new support, such as the rom, as a PR in it's own right. The other changes are independent of the rom and so shouldn't really be part of the same PR (and it could be argues that each of these should be their own PR, but that's probably being a too pedantic!!).

Thanks!

deepan02:
sure .. i would create 2 pull request

PR #96, opened 2019-08-05
n1sdp/n1sdp scp2pcc

chandnich:
These patches add n1sdp_scp2pcc module which allows scp to communicate with pcc and also modifies the n1sdp_sensor module to send a shutdown command when the temperature reaches the shutdown threshold.


CJKay:
+1 CI failure due to the coding style check failing on the static library.

PR #103, opened 2019-08-28
Ensure system PLLs on Juno are locked within MCC timeout

CJKay:
The current approach to resetting the PLLs, which is to use a pre-main
constructor function, is not infallible - if we statically allocate
enough data, eventually the time it takes for the C runtime to
initialize exceeds the amount of time that the motherboard
microcontroller gives us to ensure the system PLLs are locked.

The workaround for this is to instead wrap the core's reset handler.
This way we ensure that the PLLs have been reset before we do any time-
consuming work.

### DIFF
--- a/arch/src/Makefile
+++ b/arch/src/Makefile
@@ -11,6 +11,7 @@ include $(BS_DIR)/toolchain.mk
 BS_LIB_NAME = arch
 
 BS_LIB_SOURCES_armv7-m += armv7-m/exceptions.c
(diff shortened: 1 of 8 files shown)


CJKay:
jenkins: ok to test

CJKay:
CI+1, check failure due to coding style script rejection of the Arm Compiler 6 $Super$$ and $Sub$$ syntax.

PR #110, opened 2019-09-18
build: enhance entry guards in internal header files

etienne-lms:
Add prefix INTERNAL_ to defined macro for internal header file.
This change prevents an internal/foo.h from obscuring generic
foo.h because former and later both guards from

#ifndef FOO_H
#define FOO_H
(...)
#endif /*FOO_H*/

As internal/*.h are included with #include <internal/*.h>, INTERNAL_
looks a generic enough prefix.

### DIFF
--- a/module/cmn600/include/internal/cmn600_ccix.h
+++ b/module/cmn600/include/internal/cmn600_ccix.h
@@ -9,8 +9,8 @@
  *      CMN600 CCIX Configuration Interface
  */
 
-#ifndef CMN600_CCIX_H
-#define CMN600_CCIX_H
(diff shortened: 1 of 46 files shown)


etienne-lms:
Hello, I have a small list of small changes in SCP-firmware. Few of such small P-Rs will follow.

FYI, this is part of a port of the SCMI server from SCP-firmware into OP-TEE secure world. P-R proposals will follow...

CJKay:
Tagging @masahisak as this change touches SynQuacer, though it should be non-breaking.

CJKay:
Jenkins: add to whitelist.

PR #112, opened 2019-09-18
module clock: minor cast, export, braces fixup

etienne-lms:
Adds explicit casts where compiler reports issues.

Prevents module_ctx variable being exported outside the scope clock
module source file.

Adds braces around a conditioned a multi-line instruction block:
consider inline comment a part of the instruction block.

### DIFF
--- a/module/clock/src/mod_clock.c
+++ b/module/clock/src/mod_clock.c
@@ -33,7 +33,7 @@ struct clock_ctx {
     unsigned int dev_count;
 };
 
-struct clock_ctx module_ctx;
+static struct clock_ctx module_ctx;
 
 /*
  * Module API functions
@@ -198,9 +198,10 @@ static int clock_bind(fwk_id_t id, unsigned int round)
     if (round == 1)
         return FWK_SUCCESS;
 
-    if (!fwk_id_is_type(id, FWK_ID_TYPE_ELEMENT))
+    if (!fwk_id_is_type(id, FWK_ID_TYPE_ELEMENT)) {
         /* Only element binding is supported */
         return FWK_SUCCESS;
+    }
 
     ctx = module_ctx.dev_ctx_table + fwk_id_get_element_idx(id);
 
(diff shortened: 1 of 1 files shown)


module/clock/src/mod_clock.c
```diff
@@ -198,9 +198,10 @@ static int clock_bind(fwk_id_t id, unsigned int round)
     if (round == 1)
         return FWK_SUCCESS;
 
-    if (!fwk_id_is_type(id, FWK_ID_TYPE_ELEMENT))
+    if (!fwk_id_is_type(id, FWK_ID_TYPE_ELEMENT)) {
         /* Only element binding is supported */
         return FWK_SUCCESS;
+    }
```
souvikkc: I am not sure we wanted to put braces here for since line ifs. I would wait for comment from others.
Etienne could you please specify if your checkpatch complains here?

module/clock/src/mod_clock.c
```diff
@@ -278,10 +280,10 @@ static int clock_process_pd_pre_transition_notification(
     };
 
     pd_params = (struct mod_pd_power_state_pre_transition_notification_params *)
-        event->params;
+        (void *)event->params;
```
souvikkc: Which compiler are you using? Could you please let us know the exact compilation warning here, so we can put this in our internal checks if missing.

etienne-lms: I use a cortex-A toolchain, since building for cortex-A OP-TEE. Currently GCC 8.3:
`arm-linux-gnueabihf-gcc (GNU Toolchain for the A-profile Architecture 8.3-2019.03 (arm-rel-8.36)) 8.3.0`

CJKay: Could you paste the error? I'm unable to reproduce this with ARM64 GCC 8.2: https://gcc.godbolt.org/z/B-A-F5

module/clock/src/mod_clock.c
etienne-lms: Actually I didn't see checkpatch complaining. This is more a personal feeling. Feel free to reject the change, yet I think it makes the code simpler.

module/clock/src/mod_clock.c
etienne-lms: well, I checked back and... did not produce any build warning (nor issue).
I'll discard these changes.

module/clock/src/mod_clock.c
CJKay: This change is alright by me - the current style is incorrect.

etienne-lms: ok, i keep these changes in the proposed series.

etienne-lms:
updated: casts removed.

CJKay:
Jenkins: test this please.

CJKay:
Can you take out the "add explicit casts" part of the commit message before I merge this?

etienne-lms:
ok

CJKay:
Perfect, thanks.

PR #113, opened 2019-09-18
fwk_interrupt: add missing declaration of exported functions

etienne-lms:
Add missing declaration of exported function fwk_interrupt_init()
and fwk_interrupt_set_isr_fault().

### DIFF
--- a/framework/include/fwk_interrupt.h
+++ b/framework/include/fwk_interrupt.h
@@ -11,6 +11,7 @@
 #ifndef FWK_INTERRUPT_H
 #define FWK_INTERRUPT_H
 
+#include <fwk_arch.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <limits.h>
@@ -55,6 +56,16 @@
  *      Please refer to the individual function documentation for more details.
  */
 #define FWK_INTERRUPT_EXCEPTION (UINT_MAX - 2)
+
+/*!
+ * \brief Register interrupt driver in the framework.
+ *
+ * \param driver Interrupt driver instance to register
+ *
+ * \retval FWK_SUCCESS Operation succeeded.
+ * \retval FWK_E_PARAM One or more parameters were invalid.
+ */
+int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *driver);
 
 /*!
  * \brief Enable interrupts.
--- a/framework/include/internal/fwk_interrupt.h
+++ b/framework/include/internal/fwk_interrupt.h
@@ -1,0 +1,44 @@
+/*
+ * Arm SCP/MCP Software
+ * Copyright (c) 2015-2019, Arm Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Description:
+ *     Interrupt management internal resources.
+ */
+
+#ifndef INTERNAL_FWK_INTERRUPT_H
+#define INTERNAL_FWK_INTERRUPT_H
+
+/*!
+ * \addtogroup GroupLibFramework Framework
+ * @{
+ */
+
+/*!
+ * \defgroup GroupInterrupt Interrupt Management
+ * @{
+ */
+
+/*!
+ * \brief Set the fault interrupt service routine. This is used internally by
+ *      the framework for test purposes.
+ *
+ * \param isr Pointer to the fault interrupt service routine function.
+ *
+ * \retval FWK_SUCCESS Operation succeeded.
+ * \retval FWK_E_PARAM One or more parameters were invalid.
+ * \retval FWK_E_INIT The component has not been initialized.
+ */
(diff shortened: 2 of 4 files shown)


framework/include/fwk_interrupt.h
```diff
@@ -187,6 +196,18 @@ int fwk_interrupt_set_isr_param(unsigned int interrupt,
...
+ * \brief Set the fault interrupt service routine. This is used internally by
+ *      the framework.
+ *
+ * \param isr Pointer to the fault interrupt service routine function.
+ *
+ * \retval FWK_SUCCESS Operation succeeded.
+ * \retval FWK_E_PARAM One or more parameters were invalid.
+ * \retval FWK_E_INIT The component has not been initialized.
+ */
+int fwk_interrupt_set_isr_fault(void (*isr)(void));
```
souvikkc: Both these calls are used exclusively by the Framework only for internal purposes. I see currently we have externed them and used them. However, we should not put these declarations here, since modules also might use these files. Would you consider putting them in an internal header perhaps?

etienne-lms: Ok, I'll see how to move into internal `fwk_interrupt.h` header file.

CJKay: I think this function can be removed entirely - it isn't used anywhere except the tests, and the tests are testing... this function. It is, as far as I can see, entirely pointless, especially given the comment:

```
/* This function is only for internal use by the framework */
```

Which seems to be untrue. `fwk_arch_interrupt_driver::set_isr_fault` and the functions that use it also needs removing. It's up to the architecture layer to define proper ISR fault routines, as the framework cannot determine what is safe to do within the fault, nor which fault it even is.

framework/include/fwk_interrupt.h
```diff
@@ -56,6 +57,14 @@
  */
 #define FWK_INTERRUPT_EXCEPTION (UINT_MAX - 2)
 
+/*!
+ * \brief Initialize interrupt management in framework.
```
CJKay: `in the framework`

etienne-lms:
updated:
- fix inline comment
- declare `fwk_interrupt_set_isr_fault()`  from `internal/fwk_interrupt.h`.
(the functions needs to be declared since not `static` in `fwk_interrupt.c`)

I leave it to you to properly remove `fwk_interrupt_set_isr_fault()` and related resources.

CJKay:
Jenkins: test this please.

CJKay:
Jenkins: test this please.

framework/include/fwk_interrupt.h
```diff
@@ -56,6 +57,14 @@
  */
 #define FWK_INTERRUPT_EXCEPTION (UINT_MAX - 2)
 
+/*!
+ * \brief Initialize interrupt management in the framework.
+ *
```
CJKay: Missing `\param` for _driver

framework/include/fwk_interrupt.h
```diff
@@ -56,6 +57,14 @@
...
+ * \retval FWK_SUCCESS Operation succeeded.
+ * \retval FWK_E_PARAM One or more parameters were invalid.
+ */
+int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *_driver);
```
CJKay: `_driver` -> `driver` - the underscore exists to differentiate the static global in the `.c` file.

etienne-lms: ok, so i'll change fwk_interrupt.c line 23, ok?
```patch
 static const struct fwk_arch_interrupt_driver *driver;
 static unsigned int critical_section_nest_level;
 
-int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *_driver)
+int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *driver)
 {
     /* Validate driver by checking that all function pointers are non-null */
     if (_driver == NULL)
```

etienne-lms: discard my last comment :)

etienne-lms:
Fixed `fwk_interrupt_init()` declaration in header file.
Update `test_fwk_interrupt.c` to get `fwk_interrupt_init()` and `fwk_interrupt_set_isr_fault()` declared from the related header file.

framework/include/internal/fwk_interrupt.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+/*!
+ * \defgroup GroupInterrupt Interrupt Management
+ * @{
+ */
+
+/*!
+ * \brief Set the fault interrupt service routine. This is used internally by
+ *      the framework for test purpose.
```
CJKay: purposes

etienne-lms: done

etienne-lms:
fix inline comment as per https://github.com/ARM-software/SCP-firmware/pull/113#discussion_r331924917.

PR #114, opened 2019-09-18
log: discard non standard %e format

etienne-lms:
Prior this change was %e a valid format for printing error codes as
strings when identified. This change removes such support and replace
use of %e with use of standard format identifier and a error code
to string conversion function.

With this change, any standard printf-like format comply with
SCP-firmware traces implementation. fwk_err2str() helps traces to get
nice string identifier for errors.

### DIFF
--- a/framework/include/fwk_status.h
+++ b/framework/include/fwk_status.h
@@ -82,6 +82,15 @@
 #define FWK_E_PANIC         -18
 
 /*!
+ * \brief Return a human readable string representation of a status code.
+ *
+ * \param status Status code value.
+ *
+ * \return String representation of \p status
+ */
+const char *fwk_status_str(int status);
+
+/*!
  * @}
  */
 
--- a/framework/src/Makefile
+++ b/framework/src/Makefile
@@ -23,6 +23,7 @@ endif
 ifeq ($(BUILD_HAS_NOTIFICATION),yes)
     BS_LIB_SOURCES += fwk_notification.c
 endif
+BS_LIB_SOURCES += fwk_status.c
 
 BS_LIB_INCLUDES += $(ARCH_DIR)/include
 BS_LIB_INCLUDES += $(FWK_DIR)/include
--- a/framework/src/fwk_status.c
+++ b/framework/src/fwk_status.c
@@ -1,0 +1,52 @@
+/*
+ * Arm SCP/MCP Software
+ * Copyright (c) 2019, Arm Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Description:
+ *     Framework status code helpers.
+ */
+
+#include <fwk_macros.h>
+#include <fwk_status.h>
+
+static const char * const error_string[] = {
+    [-FWK_E_PARAM]       = "E_PARAM",
+    [-FWK_E_ALIGN]       = "E_ALIGN",
+    [-FWK_E_SIZE]        = "E_SIZE",
+    [-FWK_E_HANDLER]     = "E_HANDLER",
+    [-FWK_E_ACCESS]      = "E_ACCESS",
+    [-FWK_E_RANGE]       = "E_RANGE",
+    [-FWK_E_TIMEOUT]     = "E_TIMEOUT",
+    [-FWK_E_NOMEM]       = "E_NOMEM",
+    [-FWK_E_PWRSTATE]    = "E_PWRSTATE",
+    [-FWK_E_SUPPORT]     = "E_SUPPORT",
+    [-FWK_E_DEVICE]      = "E_DEVICE",
+    [-FWK_E_BUSY]        = "E_BUSY",
+    [-FWK_E_OS]          = "E_OS",
+    [-FWK_E_DATA]        = "E_DATA",
+    [-FWK_E_STATE]       = "E_STATE",
+    [-FWK_E_INIT]        = "E_INIT",
+    [-FWK_E_OVERWRITTEN] = "E_OVERWRITTEN",
+    [-FWK_E_PANIC]       = "E_PANIC",
+};
+
+static const char * const status_string[] = {
+    [FWK_SUCCESS]        = "SUCCESS",
+    [FWK_PENDING]        = "PENDING",
+};
+
+const char *fwk_status_str(int status)
+{
+    static const char unknown[] = "unknown";
+
+    unsigned int error_idx = (unsigned int)(-status);
+
+    if ((status < 0) && (error_idx < FWK_ARRAY_SIZE(error_string)))
+        return error_string[error_idx];
+    else if ((status >= 0) && (status < (int)FWK_ARRAY_SIZE(status_string)))
+        return status_string[status];
+
+    return unknown;
+}
--- a/module/power_domain/src/mod_power_domain.c
+++ b/module/power_domain/src/mod_power_domain.c
@@ -294,7 +294,7 @@ static const unsigned int mod_pd_cs_level_state_shift[MOD_PD_LEVEL_COUNT] = {
  * Internal variables
  */
 static struct mod_pd_ctx mod_pd_ctx;
-static const char driver_error_msg[] = "[PD] Driver error %e in %s @%d\n";
+static const char driver_error_msg[] = "[PD] Driver error %s (%d) in %s @%d\n";
 
 static const unsigned int tree_pos_level_shift[MOD_PD_LEVEL_COUNT] = {
     MOD_PD_TREE_POS_LEVEL_0_SHIFT,
@@ -711,9 +711,9 @@ static int initiate_power_state_transition(struct pd_ctx *pd)
     status = pd->driver_api->set_state(pd->driver_id, state);
 
     mod_pd_ctx.log_api->log(MOD_LOG_GROUP_DEBUG,
-        "[PD] %s: %s->%s, %e\n", fwk_module_get_name(pd->id),
+        "[PD] %s: %s->%s, %s (%d)\n", fwk_module_get_name(pd->id),
         get_state_name(pd, pd->state_requested_to_driver),
-        get_state_name(pd, state), status);
+        get_state_name(pd, state), fwk_status_str(status), status);
 
     pd->state_requested_to_driver = state;
 
@@ -1208,8 +1208,8 @@ static void process_system_shutdown_request(
 
         if (status != FWK_SUCCESS)
             mod_pd_ctx.log_api->log(MOD_LOG_GROUP_ERROR,
-                "[PD] Shutdown of %s returned %e\n",
-                fwk_module_get_name(pd_id), status);
+                "[PD] Shutdown of %s returned %s (%d)\n",
+                fwk_module_get_name(pd_id), fwk_status_str(status), status);
         else
             mod_pd_ctx.log_api->log(MOD_LOG_GROUP_DEBUG,
                 "[PD] %s shutdown\n", fwk_module_get_name(pd_id));
@@ -1795,7 +1795,7 @@ static int pd_start(fwk_id_t id)
         status = pd->driver_api->get_state(pd->driver_id, &state);
         if (status != FWK_SUCCESS) {
             mod_pd_ctx.log_api->log(MOD_LOG_GROUP_ERROR, driver_error_msg,
-                status, __func__, __LINE__);
+                fwk_status_str(status), status, __func__, __LINE__);
         } else {
             pd->requested_state = pd->state_requested_to_driver = state;
 
--- a/module/log/include/mod_log.h
+++ b/module/log/include/mod_log.h
@@ -146,7 +146,6 @@ struct mod_log_api {
      *      * \%s - string format
      *      * \%u - unsigned 32-bit decimal format
      *      * \%x - 32-bit hexadecimal format
-     *      * \%e - framework error code format
      *
      *      Numeric formats also accept a padding flag '0\<width\>' between the
      *      '\%' and the format specifier where the resulting string number will
--- a/module/log/src/mod_log.c
+++ b/module/log/src/mod_log.c
@@ -25,28 +25,6 @@ static struct mod_log_driver_api *log_driver;
                          MOD_LOG_GROUP_INFO | \
                          MOD_LOG_GROUP_WARNING)
 
-static const char * const errstr[] = {
-    [FWK_SUCCESS]        = "SUCCESS",
-    [-FWK_E_PARAM]       = "E_PARAM",
-    [-FWK_E_ALIGN]       = "E_ALIGN",
-    [-FWK_E_SIZE]        = "E_SIZE",
-    [-FWK_E_HANDLER]     = "E_HANDLER",
-    [-FWK_E_ACCESS]      = "E_ACCESS",
-    [-FWK_E_RANGE]       = "E_RANGE",
-    [-FWK_E_TIMEOUT]     = "E_TIMEOUT",
-    [-FWK_E_NOMEM]       = "E_NOMEM",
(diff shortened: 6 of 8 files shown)


etienne-lms:
Ok, thanks. I'm interested. I currently hack into module `log` to fit with OP-TEE trace support.

framework/include/fwk_status.h
```diff
@@ -81,6 +81,15 @@
...
 #define FWK_E_PANIC         -18
 
+/*!
+ * \brief Return a human readable string representation of an error number
+ *
+ * \param err Error numberevent Event to put into the queue for processing. Must not be \c NULL.
+ *
+ * \retval A valid string pointer.
+ */
+const char *fwk_err2str(int errno);
```
CJKay: Missing header prefix. `fwk_status_str`?

framework/include/fwk_status.h
```diff
@@ -81,6 +81,15 @@
 /*! Unrecoverable error */
 #define FWK_E_PANIC         -18
 
+/*!
+ * \brief Return a human readable string representation of an error number
+ *
+ * \param err Error numberevent Event to put into the queue for processing. Must not be \c NULL.
+ *
+ * \retval A valid string pointer.
```
CJKay: `\retval String representation of \p err.`

framework/src/fwk_errno.c
```diff
@@ -0,0 +1,49 @@
...
+    [-FWK_E_SUPPORT]     = "E_SUPPORT",
+    [-FWK_E_DEVICE]      = "E_DEVICE",
+    [-FWK_E_BUSY]        = "E_BUSY",
+    [-FWK_E_OS]          = "E_OS",
+    [-FWK_E_DATA]        = "E_DATA",
+    [-FWK_E_STATE]       = "E_STATE",
+    [-FWK_E_INIT]        = "E_INIT",
+    [-FWK_E_OVERWRITTEN] = "E_OVERWRITTEN",
+    [-FWK_E_PANIC]       = "E_PANIC",
+};
```
CJKay: We need to accommodate `FWK_PENDING` somehow. It might be better to use a `switch` statement.

module/power_domain/src/mod_power_domain.c
```diff
@@ -294,7 +294,7 @@ static const unsigned int mod_pd_cs_level_state_shift[MOD_PD_LEVEL_COUNT] = {
  * Internal variables
  */
 static struct mod_pd_ctx mod_pd_ctx;
-static const char driver_error_msg[] = "[PD] Driver error %e in %s @%d\n";
+static const char driver_error_msg[] = "[PD] Driver error %d/%s in %s @%d\n";
```
CJKay: I think `%s (%d)` is likely to prove a bit more readable.

CJKay:
This is helpful for us actually. Eventually we will be moving to `sprintf` so we can avoid handling formatting ourselves.

framework/src/Makefile
```diff
@@ -10,6 +10,7 @@ BS_LIB_NAME := framework
 BS_LIB_SOURCES += fwk_arch.c
 BS_LIB_SOURCES += fwk_assert.c
 BS_LIB_SOURCES += fwk_dlist.c
+BS_LIB_SOURCES += fwk_errno.c
```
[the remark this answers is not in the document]
etienne-lms: right, I will rename.

framework/src/fwk_errno.c
etienne-lms: right, thanks.

module/power_domain/src/mod_power_domain.c
etienne-lms: i'll fix all occurrences.

framework/include/fwk_status.h
```diff
@@ -82,13 +82,13 @@
 #define FWK_E_PANIC         -18
 
 /*!
- * \brief Return a human readable string representation of an error number
+ * \brief Return a human readable string representation of an status code.
```
CJKay: `a status code`

framework/src/fwk_status.c
```diff
@@ -0,0 +1,65 @@
...
+    case FWK_E_PANIC:
+        return "E_PANIC";
+    default:
+        break;
+    }
+
+    if (status < 0)
+        return "unknown-negative";
+
+    return "unknown-positive";
```
CJKay: I think it would suffice to simply:

```
    default:
        return "unknown";
```

As I expect anywhere where it actually matters is going to print out the status code anyway. Opinions?

framework/src/fwk_status.c
```diff
@@ -0,0 +1,65 @@
...
+ * Description:
+ *     Framework status code helper.
+ */
+
+#include <fwk_status.h>
+#include <fwk_macros.h>
+
+const char *fwk_status_str(int status)
+{
+    switch (status) {
```
ronald-cron-arm: What about to have two tables, one for the positive values and one for the negative/error values instead of this long switch?

framework/include/fwk_status.h
etienne-lms: thanks

framework/src/fwk_status.c
etienne-lms: you're right. this distinction is not really useful. i'll fix.

framework/src/fwk_status.c
etienne-lms: Ok. True that the array way looks better.

etienne-lms:
updated with a fixup commit. If you're fine with it, i'll squash all fixup commits.

framework/src/fwk_status.c
```diff
@@ -11,55 +11,42 @@
...
-    }
+    unsigned int error_idx = (unsigned int)(-status);
+
+    if (status >= 0 && status < FWK_ARRAY_SIZE(status_string))
+        return status_string[status];
 
-    if (status < 0)
-        return "unknown-negative";
+    if (status < 0 && error_idx < FWK_ARRAY_SIZE(error_string))
+        return error_string[error_idx];
```
CJKay: If you change `error_idx` to `(unsigned int)(-status) - 1`, you can remove the `FWK_SUCCESS` entry from `error_string`.

framework/src/fwk_status.c
```diff
@@ -11,55 +11,42 @@
...
-        break;
-    }
+    unsigned int error_idx = (unsigned int)(-status);
+
+    if (status >= 0 && status < FWK_ARRAY_SIZE(status_string))
+        return status_string[status];
 
-    if (status < 0)
-        return "unknown-negative";
+    if (status < 0 && error_idx < FWK_ARRAY_SIZE(error_string))
```
CJKay: `else if (error_idx < FWK_ARRAY_SIZE(error_string))`

framework/src/fwk_status.c
```diff
@@ -11,55 +11,42 @@
...
-    case FWK_E_OVERWRITTEN:
-        return "E_OVERWRITTEN";
-    case FWK_E_PANIC:
-        return "E_PANIC";
-    default:
-        break;
-    }
+    unsigned int error_idx = (unsigned int)(-status);
+
+    if (status >= 0 && status < FWK_ARRAY_SIZE(status_string))
```
CJKay: Conditions need to be wrapped in parentheses to make precedence clear:

`(status >= 0) && (status < FWK_ARRAY_SIZE(status_string))`

CJKay:
> 
> 
> updated with a fixup commit. If you're fine with it, i'll squash all fixup commits.

Could you squash those now? It makes it a bit easier to review.

framework/include/fwk_status.h
```diff
@@ -81,6 +81,15 @@
 /*! Unrecoverable error */
 #define FWK_E_PANIC         -18
 
+/*!
+ * \brief Return a human readable string representation of a status code.
+ *
+ * \param status Status code value.
+ *
+ * \retval String representation of \p status
```
CJKay: This should be `\return` as we use `\retval` to describe individual return values.

framework/src/fwk_status.c
```diff
@@ -0,0 +1,52 @@
+/*
+ * Arm SCP/MCP Software
+ * Copyright (c) 2019, Arm Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Description:
+ *     Framework status code helper.
```
CJKay: `helpers`

framework/src/fwk_status.c
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+#include <fwk_status.h>
+#include <fwk_macros.h>
```
CJKay: Include order:

```
#include <fwk_macros.h>
#include <fwk_status.h>
```

module/power_domain/src/mod_power_domain.c
```diff
@@ -711,9 +711,9 @@ static int initiate_power_state_transition(struct pd_ctx *pd)
     status = pd->driver_api->set_state(pd->driver_id, state);
 
     mod_pd_ctx.log_api->log(MOD_LOG_GROUP_DEBUG,
-        "[PD] %s: %s->%s, %e\n", fwk_module_get_name(pd->id),
+        "[PD] %s: %s->%s, %s (%d)\n", fwk_module_get_name(pd->id),
         get_state_name(pd, pd->state_requested_to_driver),
-        get_state_name(pd, state), status);
+        get_state_name(pd, state), fwk_err2str(status), status);
```
CJKay: Lots of instances of the old name still here:

`fwk_err2str(status)` -> `fwk_status_str(status)`

framework/include/fwk_status.h
etienne-lms: ok

framework/src/fwk_status.c
etienne-lms: thanks

framework/src/fwk_status.c
etienne-lms: right

framework/src/fwk_status.c
etienne-lms: ok

framework/src/fwk_status.c
etienne-lms: ok

framework/src/fwk_status.c
etienne-lms: Would be fine but make the table a bit gory with `FWK_xxx - 1`:
```C
static const char * const status_string[] = {
    [FWK_PENDING - 1]        = "PENDING",
};
```

Would you still prefer this `- 1` way?

module/power_domain/src/mod_power_domain.c
etienne-lms: this is want happen when pushing without testing :(
my bad... i'll fix.

framework/src/fwk_status.c
CJKay: That's a fair point. I think we can simply remove `FWK_SUCCESS` from the `error_str` and otherwise leave it how it was - nothing is going to read from index `0` so we should be fine.

etienne-lms: ok. I agree having "SUCCESS" at 2 places is not fine. Fine with index 0 return `NULL`.
I'll push an update.

framework/src/fwk_status.c
etienne-lms: actually no:
 if (status positive and index in range) then (return positive status ref.)
 if (status negative and index in range) then (return negative status ref.)
 return "unknwn" ref.

framework/src/fwk_status.c
```diff
@@ -0,0 +1,52 @@
...
+
+const char *fwk_status_str(int status)
+{
+    unsigned int error_idx = (unsigned int)(-status);
+    static const char unknown[] = "unknown";
+
+    if ((status < 0) && (error_idx < FWK_ARRAY_SIZE(error_string)))
+        return error_string[error_idx];
+
+    if ((status >= 0) && (status < (int)FWK_ARRAY_SIZE(status_string)))
```
CJKay: `else if`

framework/src/fwk_status.c
```diff
@@ -0,0 +1,52 @@
...
+
+static const char * const status_string[] = {
+    [FWK_SUCCESS]        = "SUCCESS",
+    [FWK_PENDING]        = "PENDING",
+};
+
+const char *fwk_status_str(int status)
+{
+    unsigned int error_idx = (unsigned int)(-status);
+    static const char unknown[] = "unknown";
```
CJKay: Local statics are typically placed at the top of the function separated by a newline (simply to make it clear that they are not standard local variables).

module/power_domain/src/mod_power_domain.c
CJKay: We've all done it!

framework/src/fwk_status.c
CJKay: Apologies, you're right; that's what I get for doing late Friday reviews!

framework/src/fwk_status.c
etienne-lms: may I ask why an `else` is preferred here? I don't against but I feel it adds no value.

framework/src/fwk_status.c
etienne-lms: ok

framework/src/fwk_status.c
etienne-lms: (despite this is minor stuff) I meant this:
```C
    if ((status < 0) && (error_idx < FWK_ARRAY_SIZE(error_string)))
        return error_string[error_idx];

    if ((status >= 0) && (status < (int)FWK_ARRAY_SIZE(status_string)))
        return status_string[status];

    return unknown;
```
is easier to catch than this:
```C
    if ((status < 0) && (error_idx < FWK_ARRAY_SIZE(error_string)))
        return error_string[error_idx];
    else if ((status >= 0) && (status < (int)FWK_ARRAY_SIZE(status_string)))
        return status_string[status];
    else
        return unknown;
```

framework/src/fwk_status.c
etienne-lms: addressed.

framework/src/fwk_status.c
CJKay: I'm happy to leave out `else`, but I would prefer we include specifically the `else if` as it just makes it a bit more obvious that the two conditions are tightly connected (`if not a valid error code, check if valid status code`.

etienne-lms: option1:
```C
    if ((status < 0) && (error_idx < FWK_ARRAY_SIZE(error_string)))
        return error_string[error_idx];
    else if ((status >= 0) && (status < (int)FWK_ARRAY_SIZE(status_string)))
        return status_string[status];

    return unknown;
```

option2 (misra-like):
```C
    if (status < 0) {
        if (error_idx < FWK_ARRAY_SIZE(error_string))
            return error_string[error_idx];
        else
            return unknown;
    } else {
        if (status < (int)FWK_ARRAY_SIZE(status_string)))
            return status_string[status];
        else
            return unknown;
    }
```

CJKay: The first option is preferable - there's no need to do in three conditions what can be done in two, and we aren't bound by MISRA.

Additionally, there's no need for the manual cast as `size_t` has a greater conversion rank than `int`.

So we should use:

```c
    if ((status < 0) && (error_idx < FWK_ARRAY_SIZE(error_string)))
        return error_string[error_idx];
    else if ((status >= 0) && (status < FWK_ARRAY_SIZE(status_string)))
        return status_string[status];

    return unknown;
```

etienne-lms: ok, thanks.

Regarding the cast, I think it is needed: comparing a signed and an unsigned type is likely to offend the compiler.

etienne-lms:
updated (with fixup commit squashed)

etienne-lms:
Dear reviewers, is this P-R ok?

PR #117, opened 2019-09-20
synquacer: Fix GCC 7 -Wformat-truncation error

nicola-mazzucato-arm:
GCC 7 introduces the -Wformat-truncation warning and associated
heuristics. This warning identifies potentially unexpected string
truncation issues by statically range-checking the vararg inputs to
snprintf().

For us, GCC seems to be unable to reliably identify the bounds of the
inputs (cluster_idx, core_idx) even though we have static guarantees in
place as those guarantees are made in a separate compilation unit from
the one in which the error occurs.

This patch allows the compiler to inline the cluster/core count
functions to ensure it can properly ascertain the proper bounds on the
inputs to snprintf().

Change-Id: I2d2b658368531e0e0fc5d1c0b691b61e67a106a6


nicola-mazzucato-arm:
@masahisak can you please review and approve - if possible - this patch?

masahisak:
> @masahisak can you please review and approve - if possible - this patch?

It is ok to me. Thank you!

PR #125, opened 2019-10-18
n1sdp: introduce multichip information SDS region.

pmanish87:
N1SDP can support multichip configuration wherein n1sdp boards are
connected over high speed coherent CCIX link, for now only dual-chip
is supported.

Multichip information can only be probed in SCP, use SDS region to
propagate this information to TF-A.

This patch introduces a new SDS to store multichip information and a
place holder to update this information provided by C2C module which is
responsible for probing chip-to-chip information.

Currently it is configured as single-chip system but once C2C module is
added this information will be filled dynamically.

Change-Id: I57e4aac5c32aa22f927faac4a3f91876ef34dd23
Signed-off-by: Manish Pandey <manish.pandey2@arm.com>


nicola-mazzucato-arm:
Jenkins: Please test this

ronald-cron-arm:
Jenkins: ok to test

ronald-cron-arm:
jenkins: ok to test

PR #136, opened 2019-11-08
scmi_sensor: Fix return value from response event

nicola-mazzucato-arm:
This patch fixes the status returned to scmi when
the result is provided through a response event. In the
case of a hardware failure, the agent is now correctly
informed of such state.

Change-Id: I96c09e3f63afa86bef64da81da346895ea738f22


nicola-mazzucato-arm:
jenkins: ok to test

PR #154, opened 2019-12-11
power_domain: Handle power_domain set_state error

DakeyR:
While transitioning a power domain to a particular state, we should not
expect any driver input if the driver has failed to initiate a power
domain transition.

Change-Id: If9b03221b30251f01855e220e5ef2a80f5c7ef80


souvikkc:
jenkins: ok to test

PR #157, opened 2019-12-13
juno: CDCEL clock driver uses incorrect address for I2C transaction

jimqui01:
The first four bytes of the PLL Config register are not used by Juno, so
we always skip them when reading/writing the PLL configuration. The
address was not incremented for the write after a read when setting the
PLL data in set_rate_from_index() call.

Change-Id: I094d3648168a1142cba4ee90e51aa31b3b93de88


nicola-mazzucato-arm:
Jenkins: ok to test

PR #166, opened 2019-12-20
HDLCD clock fixes

raphaelgault:
The changes introduced here correct the naming of the two HDLCD clocks exposed through SCMI so that they match expectations.
The binding between the HDLCD driver and the CDCEL driver is also corrected so that the response to events get send correctly back to the client.


jimqui01:
Jenkins: ok to test

PR #175, opened 2020-01-07
juno/reg_sensor: Change the type of the fake_register to uint64_t

raphaelgault:
The value read when doing a get_value is 64-bit so we need to store a
64-bit value in the fake register.

Change-Id: I3cf451e2659436106cc93605928fde527045ac23


nicola-mazzucato-arm:
Jenkins: ok to test

PR #183, opened 2020-01-22
juno/xrp7724: Cache voltage value to save I2C transmissions

raphaelgault:
This patch introduces a cache for the voltage value for each PSU device
in order to prevent the device from doing I2C transmissions and reading
the voltage if it is already cached or set to the requested value.

Change-Id: I812b2b442069a2f4929d801c79989aea75a265f9


jimqui01:
Jenkins: OK to test

nicola-mazzucato-arm:
Jenkins: retest this please

PR #193, opened 2020-02-27
System Info and RD-N1-Edge dual-chip support

vijayenthiran-arm:
This series of patches introduces system id information module which provides HAL interface for the registers like SID or SSC to retrieve the information about the platform. CMN-600 module is made to use system info module to retrieve the chip id information. With all these, support for RD-N1-Edge dual-chip platform is also added.


jimqui01:
jenkins: ok to test

PR #206, opened 2020-03-12
fwk/fwk_mm_alloc: panic if memory allocation failed

jimqui01:
Currently the framework will return NULL if the memory allocation
fails. As the system will ultimately fail on receiving the error
we could just panic instead, saving the memory used for the NULL
checks.

Change-Id: I731b87b7acb19d30df84936b07c76cbfe0d0726e


jimqui01:
jenkins: ok to test

PR #211, opened 2020-03-16
DVFS: Event processing returns incorrect status

jimqui01:
FWK_PENDING should be returned as FWK_SUCCESS from the
event handler.

Change-Id: Ibb56750f79d3b4f2f6ca82e655667f140563eede


jimqui01:
jenkins: ok to test

PR #225, opened 2020-03-24
Introduce the logging framework

CJKay:
This PR introduces the new logging framework, intended to replace the existing `mod_log`-based logging logic. For more information see the Doxygen documentation.

This change converts all existing platforms to this new framework. Validation has been done only on the following platforms:

- SGM-775
- SGM-776
- SGI-575
- RD-N1-E1
- Juno

These platforms have been updated, but log messages may no longer appear as before, so platform maintainers are advised to validate that their platforms behave as before:

 - N1SDP (@manojkumar-arm)
 - SynQuacer (@masahisak)


masahisak:
I have verified on SynQuacer platform and log appears same as before.
Only the difference is that banner is not appeared in debug build, but it is not an issue.

PR #230, opened 2020-04-15
Add platform support RD-Daniel Config XLR

adiang01:


jimqui01:
jenkins: ok to test

jimqui01:
This failure happens because all the patches start getting tested in CI and it fails.
For internal review to scp-firmware-v2 repo, the patches are submitted in batch of 10. In that case, the CI goes through. The failure is related to "Framework build (Host, GCC): Failed" this message.
The same patches have passed internal CI tests when submitted in batch of 10.

PR #231, opened 2020-04-15
product/rddanielxlr: disable internal log buffer

vijayenthiran-arm:
Release mode builds enables log buffering functionality with a buffer
size of 4KiB. This will result in log messages not written to the
console immediately. To avoid this in release builds, disable buffering
by setting the buffer size to 0.

Change-Id: Ie2319ae0c313a136de2fdb98c67f6a87584f2309

### DIFF
--- a/product/rddanielxlr/include/fmw_log.h
+++ b/product/rddanielxlr/include/fmw_log.h
@@ -1,0 +1,16 @@
+/*
+ * Arm SCP/MCP Software
+ * Copyright (c) 2020, Arm Limited and Contributors. All rights reserved.
+ *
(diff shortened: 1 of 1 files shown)


vijayenthiran-arm:
This patch to be merged after [Pull 230](https://github.com/ARM-software/SCP-firmware/pull/230 ).

jimqui01:
jenkins: ok to test

jimqui01:
The issue is related to PLL base address change in the new 11.10.25 FVP for SGI-575 and RD-N1-Edge platforms. The patch in this pull request is for module/cmn-rhodes which is not used on either SGI-575 or RD-N1-Edge.

PR #238, opened 2020-05-06
RFC: clock: add run reference counting

jeromebrunet:
Hi,
I'm working on platform on which several agent might be using the same clock. In this configuration, it is crucial that a call to set_state() with  MOD_CLOCK_STATE_STOPPED does not stop the clock if another agent still use the clock.

I have discussed this topic with Sudeep Holla by mail. He asked me to raise a bug but I gave it a shot a tried to implement something. 

I have tested this on Juno by hacking Linux CCF to manually send set_state() request to the SCP firmware. It seems to work as expected:
* Call to set_state() will increment or decrement the run_count depending on the requested state.
* Calls to the underlying clock driver is only made when the run count transition from 0 to 1 or 1 to 0. 

What is missing:
On start, the initial clock state is check to intitialize the run_count to 0 or 1.
This init does not support delayed response ATM. It will be added before the actual submission

But first, lets see if this proposal is going in the right direction :)

Signed-off-by: Jerome Brunet <jbrunet@baylibre.com>

### DIFF
--- a/module/clock/src/mod_clock.c
+++ b/module/clock/src/mod_clock.c
@@ -40,6 +40,9 @@ struct clock_dev_ctx {
     /* Status of the pending transition */
     unsigned int transition_pending_response_status;
 
+    /* Enable count of the clock, if applicable */
+    unsigned int run_count;
+
     /* A request is on-going */
     bool is_request_ongoing;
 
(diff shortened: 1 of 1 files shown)


jimqui01:
Hi Jerome,

We (Arm) are currently in the process of defining/implementing the SCP firmware architecture to handle multiple agents accessing shared resources. We will be addressing the issue you have with the clocks as part of this effort, and would prefer to make the solution as general as possible, not targeted at clocks or indeed any other device. 

If you would be interested in getting involved in this effort please let me know, thanks.

Jim Q.

jeromebrunet:
> Hi Jerome,
> 
> We (Arm) are currently in the process of defining/implementing the SCP firmware architecture to handle multiple agents accessing shared resources. We will be addressing the issue you have with the clocks as part of this effort, and would prefer to make the solution as general as possible, not targeted at clocks or indeed any other device.

Make sense 

> 
> If you would be interested in getting involved in this effort please let me know, thanks.
> 

Sure, how may I help ?

The product I'm working on needs multi-agent/resource sharing support for clocks and power domains at least.

Clocks are quite simple but power domains (with implementation defined states) are tougher. 
The goal of this RFC was merely to start the discussion :)

> Jim Q.

jimqui01:
Hi Jerome,

This commit 

https://github.com/ARM-software/SCP-firmware/commit/49f517c42550bfaef35511ab2191eef30942181a

has added the policy handler framework and an example for clocks. Can you make any enhancements you need and submit them. Also, if you want to add a platform policy handler for power domains that would be great.

Thanks

PR #241, opened 2020-05-21
dbg: Fix the compile error when DEBUGGER flag set

jagadeeshujja:
This patch fixes the compile error seen when DEBUGGER flag is set

Change-Id: Ie9395eabf7b849cb43c2245cea3698bf916b915d
Signed-off-by: Jagadeeesh Ujja <jagadeesh.ujja@arm.com>


jimqui01:
jenkins: ok to test

PR #242, opened 2020-05-22
Introduce CMN Booker and TC0 platform

uarif1:


jimqui01:
Please split the PR into separate requests for TC0 and CMN-booker

PR #245, opened 2020-05-22
scp/dmc620: Skip binding to DDR module if it's not present

anukou:
The assumption that a DMC module config would always have a valid
DDR PHY isn't always correct, for instance, in FVP environment(s).
Bypass binding to and configuring a DDR module if such is the case.

Change-Id: I82c7cada74bc89848dae8361a1d39565e9361c9c
Signed-off-by: Anurag Koul <anurag.koul@arm.com>


jimqui01:
jenkins: ok to test

PR #246, opened 2020-05-22
module: add support for CMN Booker interconnect

uarif1:
Add initial support for CMN Booker interconnect. 

This include configuring SAM tables, programming the nodes and sufficient support for
single chip operation. When compared to other CMN such as CMN Rhodes and
CMN600, CMN Booker has support for MTSX node type and enables coherent
read/writes through HND node.


CJKay:
Jenkins: add to whitelist.

PR #249, opened 2020-05-28
juno/build: Make Juno Debug Unit optional

jimqui01:
This patch allows the Debug Unit to be removed from the firmware
at build time if it is not required.

Change-Id: I51511ecd5b41a8ba498d6c1eb5a0cb57125c0f0b


jimqui01:
jenkins: ok to test

jimqui01:
jenkins: ok to test

PR #270, opened 2020-06-16
SCMI: Protocol and Command Identifiers

jimqui01:
This patch gathers all SCMI protocol and command identifiers into a
single file. This is required to support modules which implement
features common to all SCMI protocols such as resource permissions.

Change-Id: Id62ee9fbe8464e214b5c0d1753089eddd1787504


jimqui01:
jenkins: ok to test

PR #271, opened 2020-06-23
fwk: fwk_thread_put_event_and_wait could process NULL event

jimqui01:
The fwk_thread_put_event_and_wait processing could try and
use a NULL event if both the event queue and the ISR event queues
are empty.

Change-Id: I461a32c3040293c9c8928e60d01dc17bdefff4aa


jimqui01:
jenkins: ok to test

PR #278, opened 2020-07-02
doc: SCP Firmware Threat Model

jimqui01:
The patch publishes the current threat model for release 2.6 of the
SCP firmware. This patch also modifies some of the tools scripts
to exclude PDF files.

Change-Id: I52363325c103292585afead5a6d88c355cdb4be6


jimqui01:
jenkins:ok to test

PR #279, opened 2020-07-03
doc: Add clang-format configuration

CJKay:
This patch updates the code style documentation, adjusting it to fit
existing established but undocumented rules, as well as introducing new
ones for consistency.

It also introduces a clang-format configuration, which becomes the de
jure code style arbitrator. This configuration uses a style very close
to the existing style, with some minor tweaks in places where a single
existing style was unestablished.

Change-Id: I7a6ae117a721d7e258df01a302c5a0fb18bc3133
Signed-off-by: Chris Kay <chris.kay@arm.com>

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,70 @@
+---
+BasedOnStyle: Chromium
+AlignAfterOpenBracket: AlwaysBreak
+AlignConsecutiveAssignments: 'false'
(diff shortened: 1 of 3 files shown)


jimqui01:
The CI is complaining about the example template file. This is expected and acceptable as we don't have the bandwidth to fix teh CI for this right now.

PR #281, opened 2020-07-03
Topic/ll/rebase 03072020/scmi perf stats

lukaszluba-arm:
The performance and power statistics support.


jimqui01:
jenkins: ok to test

PR #285, opened 2020-07-06
fip: Introduce 'fip' parser module

anukou:
This module implements the parsing logic for platforms
having a FIP (Firmware Image Package) based layout for
firmware image(s) residing on a non-volatile storage, such
as a Flash memory.

As more platforms adopt a FIP based boot firmware layout,
it becomes imperative to have this parsing logic as a
common module which could be shared across platforms/products.

Change-Id: I7e558301b1bb60f93670c1e2539a9e57d6964b1d
Signed-off-by: Anurag Koul <anurag.koul@arm.com>
Co-authored-by: Patrik Berglund <patrik.berglund@arm.com>


CJKay:
Jenkins: add to whitelist.

PR #287, opened 2020-07-07
SCMI_PERF: Describe Fast Channels API should return NOT_SUPPORTED

jimqui01:
If the fast channel is not implemented for the platform the
describe_fast_channels() API should return NO_SUPPORTED.

Change-Id: I9a53037831428e47f73d8e102bec120088f6ba19


jimqui01:
jenkins: ok to test

PR #289, opened 2020-07-08
SCMI_PERF: Domain fast channel attribute should only be set if channe…

jimqui01:
…l available

The domain fast channel attribute should only be set to true if the fast
channel is available for that domain.

Change-Id: I24333d80937a83ee8272877ca418cf5d1f687f58


jimqui01:
jenkins: ok to test

PR #291, opened 2020-07-08
SCMI: Split performance domain fast channel commands

leandro-arm:
In this patch fast channels are separated in different blocks,
to allow platforms to define each fast channel command separately
Also, the fast channel memory layout has been redefined for Juno
platform

Change-Id: Ic4e51251a43a340557c3d39a3422eae6de50371d


jimqui01:
jenkins: ok to test

PR #301, opened 2020-07-29
mod_scmi_perf: Fix fast channels wrong response

Sevenarth:
The SCMI module was wrongly responding with a
SCMI_NOT_SUPPORTED error instead of a SCMI_NOT_FOUND,
when an invalid message id was sent for the command:
DESCRIBE_FASTCHANNELS. So, the modules were not
conformant to the SCMI specs.

Change-Id: I34104224815ee74eb1303ef0793e5dff99bda14a
Signed-off-by: Luca Vizzarro <Luca.Vizzarro@arm.com>


jimqui01:
jenkins: ok to test

jimqui01:
jenkins: ok to test

PR #308, opened 2020-08-06
mod_scmi: Adapt available protocols

Sevenarth:
Originally the command BASE_DISCOVER_LIST_PROTOCOLS of the base
SCMI protocol would always return all the implemented SCMI
protocols. Through the use of the resource management module,
the output can now vary according to the agent's level of
access. In this commit the PSCI agent level of access has now
been changed to the system power managament protocol only.

Change-Id: I80a9f6bb1746b62b171232b977abdb1f10d8de63
Signed-off-by: Luca Vizzarro <Luca.Vizzarro@arm.com>


CJKay:
Jenkins: Add to whitelist.

CJKay:
Jenkins: test this please.

PR #316, opened 2020-08-25
SCMI notifications/fast channels code should be conditionally built into the firmware

jimqui01:
jenkins: ok to test


jimqui01:
jenkins: ok to test

CJKay (approved):

PR #333, opened 2020-09-09
product/tc0: Fix runtime errors for tc0

uarif1:
This include:

- Changing DVFS frequencies from MHz to KHz inline with recently updated DVFS module
- Not initializing NULL elements, which is not allowed after early module context initializations
- Not setting permissions for SCMI clock, DPU requires access to SCMI clock


jimqui01 (approved):
jenkins: ok to test

jimqui01:
jenkins: ok to test

PR #336, opened 2020-09-17
product/tc0: Add pd_id for pl011 module

uarif1:
This is needed due to recently introduced changes in
mod_pl011 that perform safety check to see if the device
is powered on or off. The pl011 config is split
into ROM and RAM firmware as only RAM firmware has
power domain.

Change-Id: I877b8b3963e5a605acd8af0d771b9b61909fafaf
Signed-off-by: Usama Arif <usama.arif@arm.com>


jimqui01:
jenkins: ok to test

jimqui01 (approved):

PR #355, opened 2020-10-19
juno/xrp7724: Restore error on out of margin

nicola-mazzucato-arm:
A previous commit [1] removed the status error when the margin is out of
margin. In fact this error is required to be returned so the HAL and its logic
to handle the error can take place.

Restore the error.

[1] 7b3bf3a2fc55d4a8f1a11ea416d791652f89e5da

Change-Id: Ieeaaaf23eaea85d8bfc26f0d74281f6c6114aed3


jimqui01 (approved):
jenkins: ok to test

PR #358, opened 2020-10-21
scmi_system_power/boards: Add default timer_id and add extra check

leandro-arm:
When SCMI notifications are supported a timer is required to handle
graceful request. If this feature is not used timer_id must be
set to FWK_ID_NONE.
This patch also adds check for FWK_ID_NONE to prevent a binding
with no configuration set.

Change-Id: I25f982c2e8d686ac80dd0ebc2d9e64eac1522f3c


jimqui01:
jenkins: ok to test

PR #368, opened 2020-11-16
Juno: Disable Debug Module by default

jimqui01:
This patch disables the Juno Debug module by default to free
up some memory. The option must be re-enabled when debugging.

Change-Id: Ife7f3e2c39c12167b05e9133bf956e04c3f25508


jimqui01:
jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #369, opened 2020-11-16
tools: Display correct name for rcar's compiler

nicola-mazzucato-arm:
In tools/ci.py we build rcar platform with aarch64-none-elf-gcc, but
we display it as 'ARM'.

Correct this by replacing ARM with GCC-AArch64.

Change-Id: I7c4de86f6cd4a2a007263703edee7a7c71dfa8f6


jimqui01 (approved):
jenkins: ok to test

PR #382, opened 2020-12-07
building some SCP-firmware resources without notification/multithread/...

etienne-lms:


jimqui01 (changes requested):
DVFS can handle asynchronous requests in single-threaded mode

jimqui01:
jenkins: ok to test

nicola-mazzucato-arm (commented):
PD is already capable of handling single thread execution. or am I missing something?

etienne-lms:
deprecated

PR #383, opened 2020-12-07
[RFC] Hacking framework to build SCP-firmware in OP-TEE

etienne-lms:
This RFC shows changed we had to make in SCP-firmware framework to build it in optee_os Core. A specific BUILD_OPTEE configuration switch highlight where we needed a specific OP-TEE implementation.


etienne-lms:
deprecated

PR #384, opened 2020-12-07
[RFC] framework thread context in OP-TEE threads

etienne-lms:
This RFC shows how we currently bind an OP-TEE thread context with an SCMI thread context. This is used to allow unrelated SCMI services to nicely operate each on their own OP-TEE thread context. Actually this architecture is still being discussed internally.
This P-R is not really a RFC, rather a way to share code and show how we did.


etienne-lms:
deprecated

PR #391, opened 2020-12-21
scmi_clock: Fix return code when resource is in requested state

leandro-arm:
The SCP firmware returned an error when an agent tries to set any
resource to its current state. This takes care of the behavior
when Uboot turns ON a resource and the kernel tries to again
enable it at boot (both are same agents on same mailbox channel).

Change-Id: Ib1f094ef008d36d9979f5b69a72440ca77c1813a


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #393, opened 2020-12-26
module/cmn700: Fix non-hash rcomp enable flag check bug

iwishguo:
Range compare mode flag of non-hash regions should be checked against
NONHASH_RANGE_COMP_EN bit, not HTG_RANGE_COMP_EN bit.

Signed-off-by: Heyi Guo <guoheyi@linux.alibaba.com>


vijayenthiran-arm (approved):
Thanks for the fix.

nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):
thanks @iwishguo, nice catch

PR #395, opened 2021-01-07
fwk: only use __access__(none) with GCC 11

rossburton:
GCC 10 onwards supports read_only, write_only and read_write attributes to
__access__ and only GCC 11 (which is unreleased at this point in time) also
supports none.

To fix the build on GCC 10, only use none with GCC 11.

Change-Id: I65a258fb2fc93874fe0b6fe5e695a6df030ce3a1
Signed-off-by: Ross Burton <ross.burton@arm.com>


tarek-arm:
Jenkins: ok to test

PR #399, opened 2021-01-11
fwk_module: add debug message for module processing

iwishguo:
When SCP firmware hangs in development, it is not very easy to find
which module and which stage of the module causes the hang.

Add debug messages using cli_printf() in the module processing
framework to make it easier to find which piece of code dies.

Signed-off-by: Heyi Guo <guoheyi@linux.alibaba.com>

### DIFF
--- a/framework/src/fwk_module.c
+++ b/framework/src/fwk_module.c
@@ -187,6 +187,9 @@ static void fwk_module_init_elements(struct fwk_module_ctx *ctx)
         if (!fwk_expect(element->data != NULL))
             fwk_trap();
 
+#if BUILD_HAS_DEBUGGER
+        cli_printf(NONE, "[init] +--element idx: %u\n", i);
+#endif
         status = desc->element_init(
             element_id, element->sub_element_count, element->data);
         if (!fwk_expect(status == FWK_SUCCESS))
@@ -235,6 +238,10 @@ static void fwk_module_init_module(struct fwk_module_ctx *ctx)
         fwk_module_init_element_ctxs(ctx, elements, notification_count);
     }
 
+#if BUILD_HAS_DEBUGGER
+    cli_printf(NONE, "\n[init] module: %s\n", desc->name);
+#endif
+
     status = desc->init(ctx->id, ctx->element_count, config->data);
     if (!fwk_expect(status == FWK_SUCCESS))
         fwk_trap();
@@ -271,6 +278,9 @@ static int fwk_module_bind_elements(
          element_idx++) {
         fwk_module_ctx.bind_id =
             fwk_id_build_element_id(module_ctx->id, element_idx);
+#if BUILD_HAS_DEBUGGER
+        cli_printf(NONE, "[bind] +--element idx: %u\n", element_idx);
+#endif
         status = module->bind(fwk_module_ctx.bind_id, round);
         if (!fwk_expect(status == FWK_SUCCESS)) {
             FWK_LOG_CRIT(fwk_module_err_msg_func, status, __func__);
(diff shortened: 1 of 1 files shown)


tarek-arm (changes requested):
Can you please use FWK_LOG_TRACE for this purpose? You can change the log level using a command-line option LOG_LEVEL=TRACE. 
Also, can you please add more description to the message to make it clear. You can use __func__ in your string if that helps. 
We are using [MOD] to identify messages from this file so it is better to do the same.
If possible please reduce the number of messages to improve readability.

iwishguo:
Does FWK_LOG_TRACE depend on the serial port and log module? If it does, then the firmware may still hang silently without any last word.

CJKay:
> 
> 
> Does FWK_LOG_TRACE depend on the serial port and log module? If it does, then the firmware may still hang silently without any last word.

Both the logging API and the debugger API output are routed through the I/O API (see [here](https://github.com/ARM-software/SCP-firmware/blob/2a519a7f5b939847c05596328db2e136ce08b86c/framework/src/fwk_log.c#L243) and [here](https://github.com/ARM-software/SCP-firmware/blob/2a519a7f5b939847c05596328db2e136ce08b86c/debugger/src/cli/cli.c#L527)).

The debugger API output is always written to the I/O API's [standard output stream](https://github.com/ARM-software/SCP-firmware/blob/b8e4d2500c484e9e92e650e08047a926cfbc7727/framework/include/fwk_io.h#L500-L508), which is configured on a per-platform basis through [`FMW_IO_STDOUT_ID`](https://github.com/ARM-software/SCP-firmware/blob/b8e4d2500c484e9e92e650e08047a926cfbc7727/framework/include/fwk_io.h#L76-L88). For example, [Juno defines its standard output entity to be UART0](https://github.com/ARM-software/SCP-firmware/blob/b8e4d2500c484e9e92e650e08047a926cfbc7727/product/juno/scp_ramfw/fmw_io.h#L15).

The logging API output is written to the logging API's [log drain stream](https://github.com/ARM-software/SCP-firmware/blob/2a519a7f5b939847c05596328db2e136ce08b86c/framework/src/fwk_log.c#L37), which is configured on a per-platform basis through [`FMW_LOG_DRAIN_ID`](https://github.com/ARM-software/SCP-firmware/blob/2a519a7f5b939847c05596328db2e136ce08b86c/framework/include/fwk_log.h#L67-L78). If it hasn't been configured (which, as of today, it isn't on any platform), it defaults to the I/O API's standard output stream. So, on all platforms currently, `cli_printf()` and `FWK_LOG_X()` print in exactly the same way.

Output that goes through the I/O API - including output written using either the logging API or the debugger API - does not require its modules to have successfully initialised to be used. Instead, any module that accepts standard I/O operations can provide [an I/O adapter](https://github.com/ARM-software/SCP-firmware/blob/2a519a7f5b939847c05596328db2e136ce08b86c/framework/include/fwk_io.h#L120-L199), which allows it to be used outside of the conventional module subsystem (and, as such, before the module subsystem has started, or if it has crashed). For an example of a module that does that, [see the PL011](https://github.com/ARM-software/SCP-firmware/blob/2a519a7f5b939847c05596328db2e136ce08b86c/module/pl011/src/mod_pl011.c#L633-L639).

So, to answer your question:

> Does `FWK_LOG_TRACE` depend on the serial port?

It does not, but it may use it if your firmware is configured to do so.

> Does `FWK_LOG_TRACE` depend on the log module?

It depends only on the I/O adapter of the module, separate from the rest of the module system, and does not rely on the functioning of the scheduler (unless log buffering is enabled, but this is disabled for debug builds anyway and may optionally be disabled for release builds).

Hope that helps!

iwishguo:
Aha, it seems the log framework has changed a lot in the mainline since I last navigated the code...

CJKay:
Ah, yes, the logging machinery underwent a bit of an overhaul about half way through mid-2020. That was predominantly driven by the need to use the logging facilities outside of the module runtime, for such cases as this one. 🙂

tarek-arm:
Not active for a long time. If still required a new PR can be created.

PR #400, opened 2021-01-11
Minor fix in cpu.mk of bug causing an intended error message to not be displayed

yaronm-hailo:
When building without setting `BS_FIRMWARE_CPU` to a valid option I was not getting the intended error message because of a typo in the error line in `cpu.mk` - it should be `$(error ...)` and not `$(erro ...)`

### DIFF
--- a/tools/build_system/cpu.mk
+++ b/tools/build_system/cpu.mk
@@ -58,7 +58,7 @@ else ifeq ($(BS_FIRMWARE_CPU),host)
     BS_ARCH_VENDOR := none
     BS_ARCH_ARCH := host
 else
-    $(erro "$(BS_FIRMWARE_CPU) is not a supported CPU. Aborting...")
+    $(error "$(BS_FIRMWARE_CPU) is not a supported CPU. Aborting...")
 endif
 
 endif


nicola-mazzucato-arm:
nice catch!
would you mind renaming your patch to something like:
bs/cpu.mk: Fix typo in generating error message

That would align with our contributing guidelines
thanks

yaronm-hailo:
Did I rename it correctly?

CJKay (approved):
Well spotted!

nicola-mazzucato-arm:
Jenkins: ready to test

nicola-mazzucato-arm:
Jenkins: ok to test

yaronm-hailo:
When trying to run `make doc` with this fix applied, I get the error that `" is not a supported CPU. Aborting...`.

This is because `doc/Makefile` includes `rules.mk` which in turn includes `cpu.mk`.
Without this fix, cpu.mk would just fail quietly when running `make doc` (because of the typo this PR fixes).
It seems that the only thing `doc/Makefile` uses from `rules.mk` is the rule that creates the `build/doc` directory.

Also wanted to ask - I see the build for this PR has failed, is there a way for me to see details on the failure?

nicola-mazzucato-arm:
Hi @yaronm-hailo sorry for the delay on processing this PR. Unfortunately there's no much visibility externally of why the PR failed the test. Let me retrigger the test and I'll share with you the errors. Thanks

nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm:
@yaronm-hailo 
The PR fails the test because the proposed fix would "break" 'make doc' command. The issue is that although the proposed fix makes the syntax correct, when doing 'make doc' we get:
`/home/scp_v2/tools/build_system/cpu.mk:64: *** recipe commences before first target.  Stop.
Makefile:153: recipe for target 'doc' failed
make: *** [doc] Error 2
`
We may need a way to filter out some targets before issuing the error.
We are also gradually moving to cmake so hopefully the above case should be caught early.

tarek-arm:
Not active for a long time. If still required a new PR can be created.

PR #408, opened 2021-01-29
product/morello: Fix comparison within assert

girishpathak:
In mod_morello_pll.c, there is an assert call to check the incoming
rate which is uint64_t with UINT16_MAX*FWK_MHZ. However the resulting
value is not stored in correct data type which causes calculation to
fail. This commit fixes this calculation.

Note, this is visible only in the debug build since assert calls are
enabled only in debug builds.

Change-Id: I682882bf52c68e099d8d9959915fa26744ae845f


girishpathak:
jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #412, opened 2021-02-09
Add port aggreation support to CMN-650 and minor fixes in CMN-700

vijayenthiran-arm:
This patch series adds port aggregation support to CMN-650 and enables it in RD-V1-MC product. Also, the last two patches do minor fixes in CMN-700 module.


nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

PR #414, opened 2021-02-15
N1sdp sensor rework

arnoldgb:
Split the sensor library into SCP framework part and confidential driver as library for easier maintenance by the SCP core framework team.


tarek-arm:
Jenkins: ok to test

PR #415, opened 2021-02-17
morello: Modify morello_platform_info structure

chandnich:
In Morello platform a small portion of the memory
in bytes will be carved out for tag storage in client
mode of operation and hence the memory size variable is
changed from GB to Bytes

Signed-off-by: Manoj Kumar <manoj.kumar3@arm.com>
Signed-off-by: Chandni Cherukuri <chandni.cherukuri@arm.com>
Change-Id: I9f23d7c3d459290e102b7cc50a5bb7f5417fb6a4


tarek-arm:
Jenkins: OK to test

PR #419, opened 2021-02-28
fwk/fwk_log.c: fix erroneous check preventing log drain from working

yaronm-hailo:
The condition that ensures `fwk_log_stream` is only set if `fwk_io_open` succeeds
is wrong - it only set `fwk_log_stream` if `fwk_io_open` failed.

### DIFF
--- a/framework/src/fwk_log.c
+++ b/framework/src/fwk_log.c
@@ -55,7 +55,7 @@ int fwk_log_init(void)
         fwk_log_stream = fwk_io_stdout;
     else if (!fwk_id_is_equal(FMW_LOG_DRAIN_ID, FWK_ID_NONE)) {
         status = fwk_io_open(&stream, FMW_LOG_DRAIN_ID, FWK_IO_MODE_WRITE);
-        if (!fwk_expect(status == FWK_SUCCESS))
+        if (fwk_expect(status == FWK_SUCCESS))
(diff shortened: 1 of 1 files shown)


nicola-mazzucato-arm (commented):
Hi @yaronm-hailo , thanks for spotting this. Would you be able to provide a bit more information on how you find this issue? Many thanks

yaronm-hailo:
@nicola-mazzucato-arm I was trying to change the firmware logging interface to a different stream instead of to stdout.
To do so, I defined `FMW_LOG_DRAIN_ID` in my `fmw_io.h` to something different than `FMW_IO_STDOUT_ID`.
Then, on boot when the FW prints the first character to the logger (when writing the scp firmware logo banner), I was getting an assertion failure in `fwk_io_putch` in the check that `(stream != NULL)`.
And `fwk_log_stream` was indeed `NULL` because of the bug this patch fixes.

CJKay (approved):

tarek-arm:
Jenkins: OK to test

nicola-mazzucato-arm:
thank you @yaronm-hailo for the details. will merge soon.

nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #422, opened 2021-03-04
Topic/cmake qa v1

girishpathak:
This change is part of a larger CMake patch set, which will be introduced in the coming weeks. The main commit is the "cmake: Introduce the skeletal CMake build system" which introduces cmake skeletal build system for this project.

Note: this is just a partial commit from CMake patch set and only done so that we can verify forthcoming patches on CI.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,3 @@
----
 BasedOnStyle: Chromium
 AlignAfterOpenBracket: AlwaysBreak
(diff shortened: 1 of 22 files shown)


girishpathak:
jenkins: ok to test

nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

PR #423, opened 2021-03-04
Cmake multi tool v1

girishpathak:
This change is part of a larger CMake patch set, which will be introduced in the coming weeks. 

This PR adds some misc feature to CMake build system

Notably 
- Sets default build to release
- Adds C standards and C11 plus extensions
- Multi-toolchain support

Note: this is just a partial commit from CMake patch set and only done so that we can verify forthcoming patches on CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,23 @@ project(
     LANGUAGES NONE)
 
 #
+# Configure the default build type to be "Release". We choose to default to a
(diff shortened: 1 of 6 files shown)


girishpathak:
jenkins: ok to test

nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

PR #424, opened 2021-03-04
Cmake core v1

girishpathak:
This PR adds core CMake build support for the SCP firmware. However, note that this PR doesn't enable the building of any of the firmware targets.  Subsequent patches will be added in the coming weeks which will first support the building of the modules of the SCP firmware and later the actual firmware targets which will then finally enable the building of the platform-specific firmware binaries.  Also intended is a document that would explain how to build SCP firmware using CMake.


girishpathak:
jenkins: ok to test

nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

PR #427, opened 2021-03-09
Cmake arch target v01

girishpathak:
This PR brings futher changes which are part of overall CMake build system change for the SCP-Firmware.

Mainly, it adds.

1) 'host' build support along with stdio module build support.
2) Building of arch/arm/armv7m target.
3) And other features (Generating Map files and flat binary file generation).


nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

PR #429, opened 2021-03-11
apremap: fix 2 GiB range check

vijayenthiran-arm:
The function that checks whether the address is above 2 GiB incorrectly
checks if the address is between 1 GiB and 2 GiB. Fix this.

Change-Id: I2d2a7a785a8d098a507a650f1b0db2cc2842f56c


nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

PR #434, opened 2021-03-15
cmn700: exclude CCG devices when calculating external rnsam count

vijayenthiran-arm:
CMN-700 has new device type CCG which act as CXL gateway in addition to
CXRH, CXHA, CXRA (CCIX Gateway blocks). Exclude CCG device from the
count of devices that contain RNSAM table.

Change-Id: Ie253dcd40a7215c84a956d5978a7af8ebcc1c297


tarek-arm:
Jenkins: OK to test

nicola-mazzucato-arm (approved):

PR #436, opened 2021-03-16
cmake: product/juno

girishpathak:
This change adds CMake build system support for product/juno

Please note that firmware targets do not use libRTX_CM3.a
or RTX_CM3.lib, instead in CMake build, the library
is built using sources.

Change-Id: Iefcad29d75fc240237295295f224278254cdc9fd


tarek-arm:
Jenkins: OK to test

nicola-mazzucato-arm (approved):

PR #437, opened 2021-03-16
cmake: product/rdv1

girishpathak:
This change adds CMake build support for rdv1 platform.

Please note that firmware targets do not use libRTX_CM3.a
or RTX_CM3.lib, instead in CMake build, the library
is built using sources.

Change-Id: If5528341e9380a5c0e53b8d11ce905089ddf6bd4


nicola-mazzucato-arm:
Jenkins: ok to test

PR #444, opened 2021-03-18
scmi: fix psci protocol-count assertions

HelloGravity:
I got assertion errors when trying to work with mod_scmi with a PSCI agent with `protocol_count=0`.

It turned out the assertion itself has a bug, it does not allow `scmi_ctx.protocol_count` to be equal to `scmi_ctx.config->dis_protocol_count_psci`.

The reason for the assertion is so that the subtraction `scmi_ctx.protocol_count - scmi_ctx.config->dis_protocol_count_psci` would not be negative. But no reason it cannot be 0.

It worked after fixing these 2 assertions.


AhmedImbabyGadallah:
Hello,
I don't think it is valid to have an agent with no protocols configured for it. What is the use case for this?

tarek-arm:
Jenkins: Ok to test

PR #476, opened 2021-05-24
CMN-700 features update

vijayenthiran-arm:
This pull request adds the following three features to CMN-700 module:

- add support to configure additional non-hash mem regions
- update root node offset calculation for bigger mesh

Additionally there's a minor fix when checking the base address and size alignment.


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #479, opened 2021-05-26
Add mcp ram firmware support for RD-N2 

shriram-k-arm:
This patch series adds support for mcp ramfw for RD-N2 product


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #490, opened 2021-06-17
Rdv1 mcpramfw

shriram-k-arm:
This patch-set adds MCP RAM Firmware support for RD-V1 Platform.


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #520, opened 2021-08-23
mod_system_pll: fix incorrect return value

shriram-k-arm:
The fucntion "freq_to_half_cycle_ps" should return unsigned int values
only. But the condition that checks whether the given frequency is a
multiple of 1 KHz, will return FWK_E_PARAM which is -1. So return 0
instead of FWK_E_PARAM, if the condition fails.

Change-Id: Ib31b65757ec3824ed945a4ac5a8e4abdf860e4a7
Signed-off-by: Shriram K <shriram.k@arm.com>


tarek-arm:
Jenkins: Ok to test

nicola-mazzucato-arm (approved):

PR #522, opened 2021-08-23
Ag/voltage domain cmake fix

AhmedImbabyGadallah:
This patch fix CMake build when Voltage Domain modules are added. Also TC0 platform CMake file is updated.


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #543, opened 2021-10-04
module: Fix required MISRA defects rule 10.3

katvin01:
Further MISRA 10.3 defects will be addressed in here, relating to files within the module directory.


katvin01:
Obsolete

PR #557, opened 2021-11-22
scmi_perf: fix describe levels handler

vingu-linaro:
In case of sucess, return the payload instead of null pointer

### DIFF
--- a/module/scmi_perf/src/mod_scmi_perf.c
+++ b/module/scmi_perf/src/mod_scmi_perf.c
@@ -796,7 +796,7 @@ static int scmi_perf_describe_levels_handler(fwk_id_t service_id,
 exit:
     scmi_perf_ctx.scmi_api->respond(service_id,
         (return_values.status == SCMI_SUCCESS) ?
-            NULL : &return_values.status,
+            &return_values.status : NULL,
         (return_values.status == SCMI_SUCCESS) ?
             payload_size : sizeof(return_values.status));
 


module/scmi_perf/src/mod_scmi_perf.c
```diff
@@ -796,7 +796,7 @@ static int scmi_perf_describe_levels_handler(fwk_id_t service_id,
 exit:
     scmi_perf_ctx.scmi_api->respond(service_id,
         (return_values.status == SCMI_SUCCESS) ?
-            NULL : &return_values.status,
+            &return_values.status : NULL,
```
nicola-mazzucato-arm, line 799: Could this be simply
`&return_values.status ?`
The status is always populated, but the size of the response payload is different.
If so, then the condition at line 798 can be removed.

vingu-linaro: I have only faced the problem of return_values.status being success but null payload.
You are right and we must always provide a payload with at least the error code.
I'm going to prepare a v2

vingu-linaro: So the fix is wrong and the code is actually correct. I have been misled by a crash happening just after this code returning nil pointer to describe levels but the copy of the payload has been already done. In case of success, we don't want it to be wrongly copied so payload is NULL whereas if there is an error we want to copy the status error code.
Sorry for the inconvenience

PR #562, opened 2021-11-29
cmake: LLVM support for armv7-m

leandro-arm:


leandro-arm:
some changes are required to be done

PR #565, opened 2021-12-08
Introducing Morello SoC platform to the SCP firmware.

anukou:
This patch-series ports Morello SoC platform to the SCP firmware. It will co-exist with the already existing Morello FVP platform and is going to have its separate RAM firmware targets. ROM firmware targets, however, shall remain to be common to both the FVP and the SoC platforms.


anukou:
Cancelling this pull request as there are a few updates and some runtime fixes that were missed. A new pull request with the necessary changes will be raised.

PR #573, opened 2021-12-21
pd/juno: Simplify get state request

tarek-arm:
This patch reduces the overhead of queuing events for retrieving
the power state of a domain. It also removes a call to the
deprecated function fwk_thread_put_event_and_wait().

Signed-off-by: Tarek El-Sherbiny <tarek.el-sherbiny@arm.com>
Change-Id: Id336ce6ccf1023f37970ae562faa5a4ccedb4642


tarek-arm:
I will push an update

PR #577, opened 2021-12-22
morello: move MHU mailbox to AP non-trusted RAM for FVP

himsha01:
As per https://github.com/ARM-software/SCP-firmware/commit/9488bc401bb4e9289264368b7bb3ba702da276be#
the AP-SCP MHU mailbox memory region was moved from
SRAM to AP Non-trusted RAM for “scp_ramfw_soc” due
to the issue with unaligned accesses with MHU SRAM.

This patch is to align the same mailbox memory region
change for “scp_ramfw_fvp”.

Signed-off-by: Himanshu Sharma <Himanshu.Sharma@arm.com>
Change-Id: I8b63afc919ad83f77a47a629961659d4586aca01


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #580, opened 2022-01-07
system_power: accept the product does not support soc_wakeup_irq

masahisak:
The product does not support or intend to use soc_wakeup_irq specifies
FWK_INTERRUPT_NONE to mod_system_power_config.soc_wakeup_irq field.
This commit checks that soc_wakeup_irq is valid before calling
fwk_interrupt_disable(). Without this change, system shutdown
fails because disable_all_irqs() returns FWK_E_DEVICE.

Signed-off-by: Masahisa Kojima <masahisa.kojima@linaro.org>
Change-Id: I8b318649eec5da98093880360c3dd27f02205902

### DIFF
--- a/module/system_power/src/mod_system_power.c
+++ b/module/system_power/src/mod_system_power.c
@@ -1,6 +1,6 @@
 /*
  * Arm SCP/MCP Software
- * Copyright (c) 2015-2021, Arm Limited and Contributors. All rights reserved.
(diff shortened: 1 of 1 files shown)


nicola-mazzucato-arm (approved):
Nice one, thanks!

nicola-mazzucato-arm (changes requested):
@masahisak please update copyright to 2015-2022
thank you

nicola-mazzucato-arm (approved):

PR #587, opened 2022-01-31
cmn700: fix encoding_bits for mesh size x or y > 8

tonykn-arm:
If either the X or Y mesh size values of a CMN-700 mesh is greater than
eight, then the Node ID should be represented using a 11-bit format
with X and Y sizes encoded using 4-bits. So fix the check that
determines X and Y bit encoding size.

Signed-off-by: Tony K Nadackal <tony.nadackal@arm.com>
Change-Id: I76f8fb51953005c6b28aacd91e3ad40c728710b7


tarek-arm (approved):

tarek-arm:
Jenkins: OK test

PR #600, opened 2022-02-16
RFC request for initial work on adding device tree support for firmware build

mikel-armbb:
These patches contain an update to the firmware cmake build system that integrates a step to process device tree files in order to populate config arrays.

The library for reading .dts files has been copied from the Zephyr project, as have some of the macro generation headers.
The processing script dt_gen_config.py is based largely on the similar script in the Zephyr project, with changes to work with the SCP frameworks.

There is a single example included with this set of a arm,pl011 device implemented as a .dts configured element.
This is implemented for the product/juno/scp_ramfw.

There are two new cmake list variables used to select modules to be configured by .dts in the product Firmware.cmake :-
SCP_DT_CONFIG_MODULES_STD - for statically configuration tables.
SCP_DT_CONFIG_MODULES_DYN - for dynamically configuration tables.

When a module is selected in one of these lists the relevant config_xxx.c (e.g. config_pl011.c) file needs to be dropped from the CMakelists file for the product. 

These lists are then processed by the common framework build files to call the .dts  operations.

For the pl011 module, once selected for .dts support, an additional config_dt_pl011.c is built alongside the main pl011 common module code. This replaces the data tables from the original files that are linked via the automatic generation in cmake.

The data tables are built from macros generated by the ,dts interpretation code. 

The pl011 can be a static configuration module, but an example of using this as a dynamic module is provided to demonstrate how this mechanism will work for modules that require this configuration. 

Issues:-
1. Not yet clear what is the best way to import the code from the Zephyr project. At present this is a simple copy and where required modify. The python library itself is intended to be separated as a standalone open-source library, but the timeframe on this is not clear. Some of the code reused from the Zephyr library has Apache 2 licencing.
2. Overall layout for .dts and .yaml bindings files will need to be revised as more modules are converted. 
3. This is still a manual process - modules must be selected and appropriate .c files added / dropped.


mikel-armbb:
replaced by new v2 pull request, rebased onto latest code

PR #607, opened 2022-02-24
morello: Update morello sensor module

chandnich:
This patch series replaces the morello sensor library with the source code. Also, build option is provided to import and use a morello sensor library in place of the source code.


girishpathak (approved):

girishpathak (approved):

girishpathak:
jenkins: ok to test

PR #623, opened 2022-03-25
scp: Update firmware version to 2.10.0

brett-warren-arm:
This patch updates the SCP-firmware version to 2.10.0.


tarek-arm (approved):

nicola-mazzucato-arm (changes requested):
- readme.md file require changes as well
- I am not sure why you would like to submit the release notes again

PR #625, opened 2022-03-28
clock: Fix wrong async driver assertions

yaronm-hailo:
The assertions were checking that the status _is_ FWK_E_SUPPORT
instead of checking that it is _not_ FWK_E_SUPPORT

### DIFF
--- a/module/clock/src/clock_tree_management.c
+++ b/module/clock/src/clock_tree_management.c
@@ -158,7 +158,7 @@ int clock_management_process_state(const struct fwk_event *event)
                  * This should not be reached because. asynchronous drivers
                  * are not supported.
                  */
-                fwk_assert(status == FWK_E_SUPPORT);
+                fwk_assert(status != FWK_E_SUPPORT);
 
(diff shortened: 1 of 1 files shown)


leandro-arm:
Hi, the intended purpose of this assertion is block the execution when an asynchronous driver is used.

yaronm-hailo:
@leandro-arm Theses asserts currently test the opposite of what they are supposed to:
 they assert that the status returned by `clk_mgmt_complete_transition` is `FWK_E_SUPPORT`.
`clk_mgmt_complete_transition` returns `FWK_E_SUPPORT` on asynchronous drivers, so this means that the current behaviour is to block on synchronous drivers. My PR fixes that.

leandro-arm:
Hi @yaronm-hailo thanks for your message. Yes, you are right about it . Thanks for finding it.

tarek-arm:
Jenkins: OK to test

nicola-mazzucato-arm (approved):

PR #626, opened 2022-03-28
clock: Fix call to get_state to use driver id instead of clock id

yaronm-hailo:
When calling the clock driver's get_state API, the provied id should be
the driver element id, and not the clock's element id.
Fixed a call to get_state that was using the clock id.

### DIFF
--- a/module/clock/src/clock_tree_management.c
+++ b/module/clock/src/clock_tree_management.c
@@ -337,7 +337,7 @@ int clock_connect_tree(struct clock_ctx *module_ctx)
 
         fwk_list_push_tail(&parent->children_list, &clk->child_node);
 
-        status = clk->api->get_state(clk->id, &current_state);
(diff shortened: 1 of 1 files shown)


nicola-mazzucato-arm (changes requested):
Hi, @yaronm-hailo thanks for finding this.
Can you please make sure the indentation is correct? We suggest to use clang-format (please note that the formatter should run only on the diff, otherwise it may format the entire file).
Thanks

yaronm-hailo:
@nicola-mazzucato-arm I've ran clang-format on this file and the indentation is correct.

nicola-mazzucato-arm:
@yaronm-hailo thanks for confirming. the patch LGTM. We are currently finalising the next release and will merge this as soon as it's completed. Thank you.

nicola-mazzucato-arm (approved):

PR #631, opened 2022-04-05
Add new features for supporting OS based architecture

vingu-linaro:
This patchset adds 3 new features that will be used by new architecture and product not yet pushed.
- Remove the infinite loop from arch init so the process of events can be done somewhere else which can be the context where we receive the event or a dedicated thread context
- Add an atomic dirver structure that provide minimal atomic operation when SCP firmware runs on top of an OS which already handles interrupt controller
- Add a stop sequence that can be called before destroying the SCP firmware


nicola-mazzucato-arm (commented):
Hi @vingu-linaro can you kindly give a more descriptive name in the PR title? Thanks

vingu-linaro:
> Hi @vingu-linaro can you kindly give a more descriptive name in the PR title? Thanks

@nicola-mazzucato-arm I have updated the title to say that it is for os based architecture. Otherwise, the 3 features are different enough that it's not easy to summarize them in a single line

framework/include/internal/fwk_core.h
```diff
@@ -32,6 +32,11 @@ int __fwk_init(size_t event_count);
  */
 noreturn void __fwk_run(void);
 
+/*
+ * \brief Processing events already raised by modules and interrupt handlers.
+ */
+void __fwk_run_event(void);
```
tarek-arm, line 38: Please rename this to `void fwk_process_event_queue(void)` to make it clear.
Also, this can be the public API used externally by the framework and other modules.

framework/src/fwk_arch.c
```diff
@@ -87,6 +89,17 @@ int fwk_arch_init(const struct fwk_arch_init_driver *driver)
         return FWK_E_PANIC;
     }
 
+    /* In case firmware running under other OS context, finish processing of
+     * any raised events/interrupts and return. Else continue to process events in
+     * a forever loop.
+     */
+#if defined(BUILD_HAS_ARCH_OS)
+    __fwk_run_event();
```
tarek-arm, line 143: This becomes: `fwk_process_event_queue();`

framework/src/fwk_arch.c
```diff
@@ -87,6 +89,17 @@ int fwk_arch_init(const struct fwk_arch_init_driver *driver)
...
+    (void)fwk_log_unbuffer();
+#else
+    __fwk_run();
```
tarek-arm, line 146: And this: `__fwk_run_main_loop();`

framework/src/fwk_core.c
```diff
@@ -294,17 +294,23 @@ int __fwk_init(size_t event_count)
     return FWK_SUCCESS;
 }
 
-noreturn void __fwk_run(void)
+void __fwk_run_event(void)
```
tarek-arm, line 299: rename: `void fwk_process_event_queue(void)`

framework/src/fwk_core.c
```diff
@@ -294,17 +294,23 @@ int __fwk_init(size_t event_count)
...
+        if (!process_isr()) {
+            break;
         }
+    }
+}
 
+noreturn void __fwk_run(void)
+{
+    for (;;) {
+        (void)__fwk_run_event();
```
tarek-arm, line 315: `fwk_process_event_queue();`

tarek-arm (changes requested):
Suggested changes. Please note to update `/tools/cppcheck_suppress_list.txt`

```
diff --git a/framework/include/fwk_arch.h b/framework/include/fwk_arch.h
index 4b23fa868..01b4d99a5 100644
--- a/framework/include/fwk_arch.h
+++ b/framework/include/fwk_arch.h
@@ -272,11 +272,6 @@ int fwk_arch_deinit(void);
  */
 void fwk_arch_suspend(void);
 
-/*!
- * \brief Process an event.
- */
-void fwk_process_event(void);
-
 /*!
  * \}
  */
diff --git a/framework/include/fwk_core.h b/framework/include/fwk_core.h
index 32b7dc2cc..f41a836d2 100644
--- a/framework/include/fwk_core.h
+++ b/framework/include/fwk_core.h
@@ -67,6 +67,13 @@
              : __fwk_put_event, struct fwk_event_light * \
              : __fwk_put_event_light)(event)
 
+
+
+/*
+ * \brief Processing events already raised by modules and interrupt handlers.
+ */
+void fwk_process_event_queue(void);
+
 /*!
  * \brief Get a copy of a delayed response event.
  *
diff --git a/framework/include/internal/fwk_core.h b/framework/include/internal/fwk_core.h
index 5c3822d72..ca580cd10 100644
--- a/framework/include/internal/fwk_core.h
+++ b/framework/include/internal/fwk_core.h
@@ -25,17 +25,13 @@
 int __fwk_init(size_t event_count);
 
 /*
- * \brief Begin waiting for and processing events raised by modules and
- *      interrupt handlers.
+ * \brief Loop forever, processing events raised by modules and interrupt
+ *      handlers. This function will suspend execution if the queue is empty and
+ *      resume on an interrupt.
  *
  * \return The function does not return.
  */
-noreturn void __fwk_run(void);
-
-/*
- * \brief Processing events already raised by modules and interrupt handlers.
- */
-void __fwk_run_event(void);
+noreturn void __fwk_run_main_loop(void);
 
 /*
  * \brief Get the event being currently processed.
diff --git a/framework/src/fwk_arch.c b/framework/src/fwk_arch.c
index bde00da80..6f82df1bc 100644
--- a/framework/src/fwk_arch.c
+++ b/framework/src/fwk_arch.c
@@ -140,10 +140,10 @@ int fwk_arch_init(const struct fwk_arch_init_driver *driver)
      * a forever loop.
      */
 #if defined(BUILD_HAS_ARCH_OS)
-    __fwk_run_event();
+    fwk_process_event_queue();
     (void)fwk_log_unbuffer();
 #else
-    __fwk_run();
+    __fwk_run_main_loop();
 #endif
 
     return FWK_SUCCESS;
@@ -169,9 +169,4 @@ void fwk_arch_suspend(void)
 #if !defined(FMW_DISABLE_ARCH_SUSPEND)
     arch_suspend();
 #endif
-}
-
-void fwk_process_event(void)
-{
-    __fwk_run_event();
-}
+}
\ No newline at end of file
diff --git a/framework/src/fwk_core.c b/framework/src/fwk_core.c
index 5038a9175..4696859a3 100644
--- a/framework/src/fwk_core.c
+++ b/framework/src/fwk_core.c
@@ -296,7 +296,7 @@ int __fwk_init(size_t event_count)
     return FWK_SUCCESS;
 }
 
-void __fwk_run_event(void)
+void fwk_process_event_queue(void)
 {
     for (;;) {
         while (!fwk_list_is_empty(&ctx.event_queue)) {
@@ -309,10 +309,10 @@ void __fwk_run_event(void)
     }
 }
 
-noreturn void __fwk_run(void)
+noreturn void __fwk_run_main_loop(void)
 {
     for (;;) {
-        (void)__fwk_run_event();
+        fwk_process_event_queue();
         (void)fwk_log_unbuffer();
         fwk_arch_suspend();
     }
diff --git a/framework/test/test_fwk_core.c b/framework/test/test_fwk_core.c
index 277511f53..9a5cfbff8 100644
--- a/framework/test/test_fwk_core.c
+++ b/framework/test/test_fwk_core.c
@@ -153,7 +153,7 @@ static void test___fwk_init(void)
         &((((struct fwk_event *)(fwk_mm_calloc_val)) + 1)->slist_node));
 }
 
-static void test___fwk_run(void)
+static void test___fwk_run_main_loop(void)
 {
     int result;
     struct fwk_event *free_event, *allocated_event;
@@ -208,7 +208,7 @@ static void test___fwk_run(void)
 
     /* Event1 processing */
     if (setjmp(test_context) == FWK_SUCCESS)
-        __fwk_run();
+        __fwk_run_main_loop();
     assert(ctx->isr_event_queue.head == &(event3.slist_node));
     assert(ctx->isr_event_queue.tail == &(notification1.slist_node));
     assert(ctx->event_queue.head == &(event2.slist_node));
@@ -227,7 +227,7 @@ static void test___fwk_run(void)
 
     /* Event2 processing */
     if (setjmp(test_context) == FWK_SUCCESS)
-        __fwk_run();
+        __fwk_run_main_loop();
     assert(ctx->isr_event_queue.head == &(event3.slist_node));
     assert(ctx->isr_event_queue.tail == &(notification1.slist_node));
     assert(ctx->event_queue.head == &(allocated_event->slist_node));
@@ -246,7 +246,7 @@ static void test___fwk_run(void)
 
     /* Response to Event1 processing */
     if (setjmp(test_context) == FWK_SUCCESS)
-        __fwk_run();
+        __fwk_run_main_loop();
     assert(ctx->isr_event_queue.head == &(event3.slist_node));
     assert(ctx->isr_event_queue.tail == &(notification1.slist_node));
     assert(fwk_list_is_empty(&ctx->event_queue));
@@ -266,7 +266,7 @@ static void test___fwk_run(void)
 
     /* Extract ISR Event3 and process it */
     if (setjmp(test_context) == FWK_SUCCESS)
-        __fwk_run();
+        __fwk_run_main_loop();
     assert(ctx->isr_event_queue.head == &(notification1.slist_node));
     assert(ctx->isr_event_queue.tail == &(notification1.slist_node));
     assert(fwk_list_is_empty(&ctx->event_queue));
@@ -286,7 +286,7 @@ static void test___fwk_run(void)
     fwk_list_push_tail(&ctx->free_event_queue, &(allocated_event->slist_node));
     free_event_queue_break = true;
     if (setjmp(test_context) == FWK_SUCCESS)
-        __fwk_run();
+        __fwk_run_main_loop();
     assert(fwk_list_is_empty(&ctx->isr_event_queue));
     assert(ctx->event_queue.head == &(allocated_event->slist_node));
     assert(ctx->event_queue.tail == &(allocated_event->slist_node));
@@ -303,7 +303,7 @@ static void test___fwk_run(void)
 
     /* Process response to Notification1 */
     if (setjmp(test_context) == FWK_SUCCESS)
-        __fwk_run();
+        __fwk_run_main_loop();
     assert(fwk_list_is_empty(&ctx->isr_event_queue));
     assert(fwk_list_is_empty(&ctx->event_queue));
 
@@ -512,7 +512,7 @@ static void test___fwk_put_notification(void)
 
 static const struct fwk_test_case_desc test_case_table[] = {
     FWK_TEST_CASE(test___fwk_init),
-    FWK_TEST_CASE(test___fwk_run),
+    FWK_TEST_CASE(test___fwk_run_main_loop),
     FWK_TEST_CASE(test_fwk_put_event),
     FWK_TEST_CASE(test_fwk_put_event_light),
     FWK_TEST_CASE(test___fwk_put_notification)
diff --git a/framework/test/test_fwk_module.c b/framework/test/test_fwk_module.c
index 98923fdd7..bedbdb831 100644
--- a/framework/test/test_fwk_module.c
+++ b/framework/test/test_fwk_module.c
@@ -200,7 +200,7 @@ int __wrap___fwk_init(size_t event_count)
     return init_return_val;
 }
 
-void __wrap___fwk_run(void)
+void __wrap___fwk_run_main_loop(void)
 {
 }
 
diff --git a/tools/cppcheck_suppress_list.txt b/tools/cppcheck_suppress_list.txt
index 17fe4cb22..910ba40f3 100755
--- a/tools/cppcheck_suppress_list.txt
+++ b/tools/cppcheck_suppress_list.txt
@@ -96,7 +96,7 @@ zerodivcond:*product/juno/module/juno_cdcel937/src/mod_juno_cdcel937.c:414
 syntaxError:*arch/arm/armv8-a/include/arch_system.h:19
 
 // Cppcheck is not able to parse returned boolean values inside if conditions
-internalAstError:*framework/src/fwk_core.c:300
+internalAstError:*framework/src/fwk_core.c:302
 
 // These assignments are used for testing
 constArgument:*framework/test/test_fwk_macros.c
```

CMakeLists.txt
```diff
@@ -331,6 +331,11 @@ set(SCP_ENABLE_OVERRIDE_FIRMWARE_NAME
     ${SCP_ENABLE_OVERRIDE_FIRMWARE_NAME_INIT}
     CACHE STRING "Override firmware binary name")
 
+cmake_dependent_option(
+    SCP_ARCHITECTURE_OS "Enable the OS execution environnemen support?"
```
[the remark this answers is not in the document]
vingu-linaro: Ok, let use SCP_SUB_SYSTEM  and I will add a section to describe this mode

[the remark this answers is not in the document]
vingu-linaro: should we use SCP_ENABLE_SUB_SYSTEM to be aligned with other SCP_ flags

vingu-linaro:
The last update has been rebased on master

tarek-arm (commented):
**_"fwk: Add atomic operation when interrupt ones are not used"_**
I would like to understand why can't the existing `struct fwk_arch_interrupt_driver` be reused to implement the functionality provided by the new 3 functions.
`fwk_interrupt_context()` ==> `fwk_interrupt_get_current`
`fwk_interrupt_lock()` => `fwk_interrupt_disable`
`fwk_interrupt_unlock()` => `fwk_interrupt_enable`

Everything else in the structure can just return success.

vingu-linaro:
> **_"fwk: Add atomic operation when interrupt ones are not used"_** I would like to understand why can't the existing `struct fwk_arch_interrupt_driver` be reused to implement the functionality provided by the new 3 functions. `fwk_interrupt_context()` ==> `fwk_interrupt_get_current` `fwk_interrupt_lock()` => `fwk_interrupt_disable` `fwk_interrupt_unlock()` => `fwk_interrupt_enable`
> 
> Everything else in the structure can just return success.

1 - fwk_interrupt_disable/enable can't be chained as they don't return previous state. OS usually returns a cookie that saves what was the status before disabling. Typically zephyr has  irq_lock() which returns a cookie that is then provided to the unlock and linux has spin_lock_irq_save. Also I'm sure that current fwk_interrupt_disable() would work on smp system because irq and thread could run on different core although SCP is still mono threaded
2 - fwk_interrupt_enable(interrupt) is meant to enable/disable a specific interrupt whereas we don't have any specific interrupt in mind in this case

tarek-arm:
> 1 - fwk_interrupt_disable/enable can't be chained as they don't return previous state. OS usually returns a cookie that saves what was the status before disabling. Typically zephyr has irq_lock() which returns a cookie that is then provided to the unlock and linux has spin_lock_irq_save. Also I'm sure that current fwk_interrupt_disable() would work on smp system because irq and thread could run on different core although SCP is still mono threaded

Are you trying to protect against recursive calls to disable/enable interrupts? 
If so, we have this protection [here](https://github.com/ARM-software/SCP-firmware/blob/master/framework/include/fwk_interrupt.h#L89)  `critical_section_nest_level`  which will track the number of disable calls and will only allow enabling the interrupts after the same number of enabling calls are invoked.

> 2 - fwk_interrupt_enable(interrupt) is meant to enable/disable a specific interrupt whereas we don't have any specific interrupt in mind in this case

Sorry I meant `fwk_interrupt_get_current`. We are using it to check if we are in an interrupt context or not [like so](https://github.com/ARM-software/SCP-firmware/blob/master/framework/src/fwk_core.c#L141)

vingu-linaro:
> > 1 - fwk_interrupt_disable/enable can't be chained as they don't return previous state. OS usually returns a cookie that saves what was the status before disabling. Typically zephyr has irq_lock() which returns a cookie that is then provided to the unlock and linux has spin_lock_irq_save. Also I'm sure that current fwk_interrupt_disable() would work on smp system because irq and thread could run on different core although SCP is still mono threaded
> 
> Are you trying to protect against recursive calls to disable/enable interrupts? If so, we have this protection [here](https://github.com/ARM-software/SCP-firmware/blob/master/framework/include/fwk_interrupt.h#L89) `critical_section_nest_level` which will track the number of disable calls and will only allow enabling the interrupts after the same number of enabling calls are invoked.

I have seen this refcount in fwk_interrupt_global_disable() but I'm not sure that this can be used when SCP is included as part of an OS. The fact that critical_section_nest_level is tested and modified without being protected makes me to wonder if this works in any case and any system

> 
> > 2 - fwk_interrupt_enable(interrupt) is meant to enable/disable a specific interrupt whereas we don't have any specific interrupt in mind in this case
> 
> Sorry I meant `fwk_interrupt_get_current`. We are using it to check if we are in an interrupt context or not [like so](https://github.com/ARM-software/SCP-firmware/blob/master/framework/src/fwk_core.c#L141)

tarek-arm:
I did not understand why the existing solution will not work. If you think it will not work for any system then let's fix the problem! 
Why do we need to have 2 functions for the same purpose? One is problematic and the other only works for certain environments!
Or maybe I don't understand the problem you are trying to solve. Can you please clarify this?
And if you think we have to have 2 different structures please document clearly the difference between each one.

vingu-linaro:
> I did not understand why the existing solution will not work. If you think it will not work for any system then let's fix the problem! Why do we need to have 2 functions for the same purpose? One is problematic and the other only works for certain environments! Or maybe I don't understand the problem you are trying to solve. Can you please clarify this? And if you think we have to have 2 different structures please document clearly the difference between each one.

I came to adding a different atomic ops structure because :
- fwk_arch_interrupt_driver seems to be tied to manage an interrupt controller which is not the case for us and it would have been a kind of abuse of some interrupt interfaces. As an example, the interrupt parameter will never be updated because we don't get the interrupt number. When SCP is a sub system of a larger SW, interrupt are managed by the OS.
- We only need 3 atomic operation to manipulate critical resource like linked list
- In zephyr case, and more generally speaking, such atomic lock function returns a value that is reused for unlocking and restoring previous state. That's why I came to the interface: flags = irq_lock(), irq_unlock(flags) and irq_context()
- SCP can be called in a context which is already irq_lock so we must not break this and restore same state.
- I haven't studied in details the behavior of fwk_interrupt_global_disable/enable in the context of baremetal cortexM system but I see some possible problem with the fact that critical_section_nest_level is Read-Modified-Write without being protected. It means it can be preempted in the middle of the modification and the content of critical_section_nest_level becomes unknown.
- By default, fwk_interrupt_lock is a wrap of fwk_interrupt_global_disable so it's not different but an OS can provide its own wrapper based on its critical section management section
- fwk_interrupt_global_disable() seems to assume that SCP can disable all interrupts but it's not always possible when running with OS or on SMP system

I have replaced fwk_interrupt_global_disable() by fwk_interrupt_lock() only in the framework part and not in modules and products because I wanted to make sure that the interface is ok but I can move all modules and product to  fwk_interrupt_lock() which by default call fwk_interrupt_global_disable

tarek-arm:
You are correct, `critical_section_nest_level` will not protect against nested interrupts. Its purpose is to protect against multiple enable/disable calls within the same context. For example in the interrupt context:
```
fwk_interrupt_global_disable()
func() {
            fwk_interrupt_global_disable()
            fwk_interrupt_global_enable()
}
fwk_interrupt_global_enable()
```
The first enable func() will not enable the interrupts.

tarek-arm:
I suggest to create a new entry in `fwk_arch_interrupt_driver` for `get_interrupt_context`. The implementation for bare-metal can be the same as the current fwk_interrupt_get_current. Change the name to is_interrupt_context so the function pointer is something like:
```
int (*is_interrupt_context)( bool *is_interrupt);
```
zephyr  irq_lock() and irq_unlock() should be called in the new arch `arch/none/zephyr/src/arch_main.c`  in the `global_enable(void)` and `global_disable(void)` implementation which is part of the `fwk_arch_interrupt_driver` structure.
Remove the `fwk_arch_atomic_driver` structure.

If this is not going to work, I am happy to discuss it further.

vingu-linaro:
> I suggest to create a new entry in `fwk_arch_interrupt_driver` for `get_interrupt_context`. The implementation for bare-metal can be the same as the current fwk_interrupt_get_current. Change the name to is_interrupt_context so the function pointer is something like:
> 
> ```
> int (*is_interrupt_context)( bool *is_interrupt);

Why do you add  a bool parameter instead of using the returned value of the function ?

> ```
> 
> zephyr irq_lock() and irq_unlock() should be called in the new arch `arch/none/zephyr/src/arch_main.c` in the `global_enable(void)` and `global_disable(void)` implementation which is part of the `fwk_arch_interrupt_driver` structure. Remove the `fwk_arch_atomic_driver` structure.
> 
> If this is not going to work, I am happy to discuss it further.

zephyr irq_unlock(flags) has a parameter which doesn't match with global_enable(void).
Then, global_enable/disable are not used to protect the linked list as fwk_interrupt_global_enable/disable() use arch_interrupts_enable/disable()

Could I suggest to also add in fwk_arch_interrupt_driver: 

int (*interrupt_lock)( unsigned int *flags);
int (*interrupt_unlock)( unsigned int flags);

tarek-arm:
We don't want to compromise on performance and that's why `fwk_interrupt_global_disable` and `arch_interrupts_disable` are inline.
We can change the parameters if that helps. I suggest this change:
```
diff --git a/arch/arm/armv7-m/include/arch_helpers.h b/arch/arm/armv7-m/include/arch_helpers.h
index 050cea7a5..a9242fb79 100644
--- a/arch/arm/armv7-m/include/arch_helpers.h
+++ b/arch/arm/armv7-m/include/arch_helpers.h
@@ -14,7 +14,7 @@
  * \note inline is necessary as this call can be used in performance sensitive
  *     path
  */
-inline static void arch_interrupts_enable(void)
+inline static void arch_interrupts_enable(unsigned int not_used)
 {
     __asm volatile("cpsie i" : : : "memory");
 }
@@ -25,9 +25,10 @@ inline static void arch_interrupts_enable(void)
  * \note inline is necessary as this call can be used in performance sensitive
  *     path
  */
-inline static void arch_interrupts_disable(void)
+inline static unsigned int arch_interrupts_disable(void)
 {
     __asm volatile("cpsid i" : : : "memory");
+    retrun 0;
 }
 
 /*!
diff --git a/framework/include/fwk_arch.h b/framework/include/fwk_arch.h
index 1c7a4b720..2361faf57 100644
--- a/framework/include/fwk_arch.h
+++ b/framework/include/fwk_arch.h
@@ -178,6 +178,14 @@ struct fwk_arch_interrupt_driver {
      * \retval ::FWK_E_STATE An interrupt is not currently being serviced.
      */
     int (*get_current)(unsigned int *interrupt);
+
+    /*!
+     * \brief Check if in interrupt context.
+     *
+     * \retval :: \c true if in an interrupt context.
+     * \retval :: \c false not in an interrupt context.
+     */
+    bool (*is_interrupt_context)(void);
 };
 
 /*!
diff --git a/framework/include/fwk_interrupt.h b/framework/include/fwk_interrupt.h
index e75050f7f..cf14ce692 100644
--- a/framework/include/fwk_interrupt.h
+++ b/framework/include/fwk_interrupt.h
@@ -84,7 +84,7 @@ int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *driver);
  * \brief Enable interrupts.
 
  */
-inline static void fwk_interrupt_global_enable(void)
+inline static void fwk_interrupt_global_enable(unsigned int flags)
 {
     /* Decrement critical_section_nest_level only if in critical section */
     if (critical_section_nest_level > 0) {
@@ -93,21 +93,22 @@ inline static void fwk_interrupt_global_enable(void)
 
     /* Enable interrupts globally if now outside critical section */
     if (critical_section_nest_level == 0) {
-        arch_interrupts_enable();
+        arch_interrupts_enable(flags);
     }
 }
 
 /*!
  * \brief Disable interrupts.
  */
-inline static void fwk_interrupt_global_disable(void)
+inline static unsigned int fwk_interrupt_global_disable(void)
 {
     critical_section_nest_level++;
 
     /* If now in outer-most critical section, disable interrupts globally */
     if (critical_section_nest_level == 1) {
-        arch_interrupts_disable();
+        return arch_interrupts_disable();
     }
+    return 0;
 }
 
 /*!
```

vingu-linaro:
I have pushed a new version which takes into account comments and outcome of the discussions

framework/include/fwk_interrupt.h
```diff
@@ -84,30 +78,17 @@ int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *driver);
  * \brief Enable interrupts.
 
  */
-inline static void fwk_interrupt_global_enable(void)
+inline static void fwk_interrupt_global_enable(unsigned int flags)
```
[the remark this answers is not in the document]
vingu-linaro: will do

framework/include/fwk_arch.h
```diff
@@ -178,6 +178,14 @@ struct fwk_arch_interrupt_driver {
      * \retval ::FWK_E_STATE An interrupt is not currently being serviced.
      */
     int (*get_current)(unsigned int *interrupt);
+
+    /*!
+     * \brief Check if in interrupt context.
+     *
+     * \retval :: \c true if in an interrupt context.
```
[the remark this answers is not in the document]
vingu-linaro: will do
check_doc.py didn't raise anything on my laptop

framework/include/fwk_interrupt.h
```diff
@@ -84,30 +78,17 @@ int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *driver);
...
-        arch_interrupts_enable();
-    }
+    arch_interrupts_enable(flags);
 }
 
 /*!
  * \brief Disable interrupts.
  */
-inline static void fwk_interrupt_global_disable(void)
+inline static unsigned int fwk_interrupt_global_disable(void)
```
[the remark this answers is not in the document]
vingu-linaro: will do

framework/include/fwk_interrupt.h
```diff
@@ -84,30 +78,17 @@ int fwk_interrupt_init(const struct fwk_arch_interrupt_driver *driver);
  * \brief Enable interrupts.
 
  */
-inline static void fwk_interrupt_global_enable(void)
+inline static void fwk_interrupt_global_enable(unsigned int flags)
 {
```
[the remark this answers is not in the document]
vingu-linaro: will do

arch/arm/armv7-m/include/arch_helpers.h
```diff
@@ -8,15 +8,29 @@
 #ifndef ARCH_HELPERS_H
 #define ARCH_HELPERS_H
 
+/*
+ * This variable is used to ensure spurious nested calls won't
+ * enable interrupts. This is been defined in fwk_interrupt.c
+ */
```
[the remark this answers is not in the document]
vingu-linaro: will do

framework/test/test_fwk_notification.c
```diff
@@ -85,6 +85,11 @@ int __wrap_fwk_interrupt_get_current(unsigned int *interrupt)
     return interrupt_get_current_return_val;
 }
 
+bool __wrap_fwk_is_interrupt_context(void)
```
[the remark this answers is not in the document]
vingu-linaro: I have updated the comment to mention that this is used by arm architecture

framework/test/test_fwk_core.c
```diff
@@ -97,6 +97,11 @@ int __wrap_fwk_interrupt_get_current(unsigned int *interrupt)
     return interrupt_get_current_return_val;
 }
```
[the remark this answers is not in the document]
vingu-linaro: ok

framework/test/fwk_test.c
```diff
@@ -15,6 +15,12 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+/*
+ * This variable is used to ensure spurious nested calls won't
+ * enable interrupts. This is been accessed from inline function defined in
+ * fwk_interrupt.h
```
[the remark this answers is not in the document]
vingu-linaro: I have updated the comment to mention that this is used by arm architecture

arch/arm/armv8-a/include/arch_helpers.h
```diff
@@ -549,15 +549,29 @@ static inline uint64_t el_implemented(unsigned int el)
 #define read_cpacr() read_cpacr_el1()
 #define write_cpacr(_v) write_cpacr_el1(_v)
 
+/*
+ * This variable is used to ensure spurious nested calls won't
+ * enable interrupts. This is been defined in fwk_interrupt.c
```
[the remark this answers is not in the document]
vingu-linaro: ok

arch/arm/armv7-m/src/arch_main.c
```diff
@@ -18,6 +18,13 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+/*
+ * This variable is used to ensure spurious nested calls won't
+ * enable interrupts. This is been accessed from inline function defined in
+ * fwk_interrupt.h
```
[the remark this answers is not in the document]
vingu-linaro: ok

framework/include/fwk_arch.h
```diff
@@ -222,6 +222,15 @@ struct fwk_arch_init_driver {
  */
 int fwk_arch_init(const struct fwk_arch_init_driver *driver);
 
+/*!
+ * \brief Stop the framework library.
+ *
+ * \retval ::FWK_SUCCESS Operation succeeded.
+ * \retval ::FWK_E_PARAM One or more parameters were invalid.
```
tarek-arm, line 229: The function does not take parameters. Please remove this line.

framework/include/fwk_arch.h
```diff
@@ -222,6 +222,15 @@ struct fwk_arch_init_driver {
...
+ * \retval ::FWK_E_PANIC Unrecoverable initialization error.
```
tarek-arm, line 230: This is not doing the initialization. You can say `Unrecoverable` or `generic` error.

framework/src/fwk_arch.c
```diff
@@ -87,6 +89,18 @@ int fwk_arch_init(const struct fwk_arch_init_driver *driver)
...
     }
 
+    /*
+     * In case firmware running under other OS context, finish processing of
+     * any raised events/interrupts and return. Else continue to process events
+     * in a forever loop.
+     */
+#if defined(BUILD_HAS_SUB_SYSTEM_MODE)
+    fwk_process_event_queue();
+    (void)fwk_log_unbuffer();
```
leandro-arm, line 99: It would be better to wrap this all in a single function like it is being done with `__fwk_run_main_loop()` to improve readability

arch/arm/armv7-m/src/arch_main.c
```diff
@@ -18,6 +18,13 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+/*
+ * This variable is used to ensure spurious nested calls won't
+ * enable interrupts. This is been accessed from inline function defined in
+ * arch_helpers.h
+ */
+unsigned int critical_section_nest_level;
```
leandro-arm, line 26: I think it should be `volatile`

arch/none/host/src/arch_main.c
```diff
@@ -35,3 +35,8 @@ int main(void)
     if (status != FWK_SUCCESS)
         panic();
 }
+
+int scmi_arch_deinit(void)
+{
+    return fwk_arch_deinit();
+}
```
tarek-arm, line 42: This is dead code we can remove it.
Also, we usually use `scmi_` for protocol related functions. It is better not to use it here.

tarek-arm (changes requested):

leandro-arm (changes requested):

product/juno/module/juno_dmc400/src/mod_juno_dmc400.c
```diff
@@ -996,14 +998,14 @@ static int juno_dmc400_start(fwk_id_t id)
         return status;
     }
 
-    fwk_interrupt_global_disable();
+    fags = fwk_interrupt_global_disable();
```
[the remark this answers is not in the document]
vingu-linaro: sorry, i'm going to fix it

framework/include/fwk_arch.h
vingu-linaro: ok

arch/none/host/src/arch_main.c
vingu-linaro: this is used after reworking host main loop. will move it in the right commit

framework/src/fwk_arch.c
vingu-linaro: With BUILD_HAS_SUB_SYSTEM_MODE, fwk_process_event_queue(), fwk_log_unbuffer() and fwk_arch_suspend() can be used in different place. That's why they are kept separated

framework/include/fwk_arch.h
vingu-linaro: yes, forgot to clean the copy/paste of fwk_arch_init

product/juno/module/juno_dmc400/src/mod_juno_dmc400.c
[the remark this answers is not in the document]
vingu-linaro: also fixed module statistics which had same problem

framework/src/fwk_module.c
```diff
@@ -449,6 +450,86 @@ int fwk_module_start(void)
...
+    int status;
+    const struct fwk_module *module;
+    unsigned int element_idx;
+
+    module = module_ctx->desc;
+
+    for (element_idx = 0; element_idx < module_ctx->element_count;
+         element_idx++) {
+
+        if (module->stop != NULL) {
```
tarek-arm, line 464: There is no need to do this check for each element. In fact, you only need to check this once for the module and don't call the function `fwk_module_stop_elements` if there is no `module->stop` defined.

framework/src/fwk_module.c
```diff
@@ -449,6 +450,86 @@ int fwk_module_start(void)
...
+        status = module->stop(module_ctx->id);
+        if (!fwk_expect(status == FWK_SUCCESS)) {
+            FWK_LOG_CRIT(fwk_module_err_msg_func, status, __func__);
+            return status;
+        }
+    }
+
+    module_ctx->state = FWK_MODULE_STATE_SUSPENDED;
+
+    return fwk_module_stop_elements(module_ctx);
```
tarek-arm, line 497: Do not call this function if `module->stop == NULL`

framework/include/fwk_arch.h
```diff
@@ -222,6 +222,14 @@ struct fwk_arch_init_driver {
  */
 int fwk_arch_init(const struct fwk_arch_init_driver *driver);
 
+/*!
+ * \brief Stop the framework library.
+ *
```
tarek-arm, line 227: please add some comments about when is this function expected to be called

framework/include/fwk_module.h
```diff
@@ -222,6 +222,27 @@ struct fwk_module {
...
+     * \details This function is called by the framework for the module and then
+     *      for all of its elements during the stop stage. Elements are
+     *      stopped in the order they are declared in the module configuration
+     *      data.
+     *
+     *      The framework does not mandate a particular purpose for this
+     *      function. It may be used to perform any final processing of the
+     *      module and its elements before entering the suspend phase.
+     *
+     * \note This function is \b optional.
```
tarek-arm, line 237: Please add a note that this is not to stop/start at runtime but rather for terminating the application.

tarek-arm (changes requested):
Can you please provide more information in both the commit and the code headers comments?

For the commit message, using the word server is confusing to anyone external.
We can say something like "When the scp is running as an application or sub-system there is a need to terminate the process gracefully. This patch introduces the deinit and stop sequence to allow the framework and each individual module to cleanup and release resources before terminating the process"

The header files comment for the `stop` and `deinit` functions must state clearly this is not to be used at runtime to stop and start components at any moment. This is only for termination and the framework is not expected to run the initialization sequence after a stop function.

framework/src/fwk_module.c
vingu-linaro: I have reused the same skeleton as fwk_module_start_module and fwk_module_start_elements which also do the check at both level. I wanted to stay consistent. Should I change both ?

framework/src/fwk_module.c
vingu-linaro: as said above, I have followed fwk_module_start_module skeleton

framework/include/fwk_module.h
vingu-linaro: ok will add:
This function is \b optional and aims to release software resources or to set hardware resources in a proper state before terminating the SCP software

framework/src/fwk_module.c
tarek-arm: @vingu-linaro  it's up to you we can do the optimization now or later if you want.

vingu-linaro: ok. So i'm going to add a patch to remove the useless check in start and will align the stop sequence

vingu-linaro: After looking a bit more at the code, don't we have to loop on each element anyway in order to update its state ? This would mean that we have to call fwk_module_start/stop_elements for each elements even if module->stop/start == NULL in order to update the state.
So I'm going to let the code as it is for now because the change doesn't seem straightforward

tarek-arm: That's fine by me

framework/src/fwk_arch.c
leandro-arm: I know but it could be better if you wrap this two in a function like

```suggestion
void __fwk_run_main_single(void)
{
    fwk_process_event_queue();
    (void)fwk_log_unbuffer();
}
```

tarek-arm:
Jenkins: OK to test

arch/arm/armv7-m/include/arch_helpers.h
```diff
@@ -8,15 +8,29 @@
...
 {
-    __asm volatile("cpsie i" : : : "memory");
+   /* Decrement critical_section_nest_level only if in critical section */
+    if (critical_section_nest_level > 0) {
+        critical_section_nest_level--;
+    }
+
+    /* Enable interrupts globally if now outside critical section */
+    if (critical_section_nest_level == 0) {
+        __asm volatile("cpsie i" : : : "memory");
```
tarek-arm, line 32: Using asm here is triggering MISRA violation. I understand this is something introduced by this patch but we have to resolve it.
I suggest using the CMSIS equivalent. Can you please replace `__asm volatile("cpsie i" : : : "memory");` with `__enable_irq();` and `__asm volatile("cpsid i" : : : "memory");` with `__disable_irq()` so we can pass this check?

tarek-arm (changes requested):

framework/src/fwk_module.c
```diff
@@ -449,6 +450,86 @@ int fwk_module_start(void)
...
+    int status;
+
+    if (!fwk_module_ctx.initialized) {
+        FWK_LOG_CRIT(fwk_module_err_msg_func, FWK_E_STATE, __func__);
+        return FWK_E_STATE;
+    }
+
+    fwk_module_ctx.stage = MODULE_STAGE_STOP;
+    status = stop_modules();
+    if (status != FWK_SUCCESS)
```
tarek-arm, line 527: We need the brackets around to body of the if statement to avoid MISRA violation.
"The body of the "then" branch of the "if" statement is not a compound statement."

framework/src/fwk_arch.c
```diff
@@ -104,6 +104,17 @@ int fwk_arch_init(const struct fwk_arch_init_driver *driver)
     return FWK_SUCCESS;
 }
 
+int fwk_arch_deinit(void)
+{
+    int status;
+
+    status = fwk_module_stop();
+    if (!fwk_expect(status == FWK_SUCCESS))
+        return FWK_E_PANIC;
```
tarek-arm, line 113: Same here. We must have the brackets.

framework/src/fwk_module.c
```diff
@@ -449,6 +450,86 @@ int fwk_module_start(void)
...
+{
+    int status;
+    unsigned int module_idx;
+    struct fwk_module_context *module_ctx;
+
+    for (module_idx = 0; module_idx < FWK_MODULE_IDX_COUNT; module_idx++) {
+        module_ctx = &fwk_module_ctx.module_ctx_table[module_idx];
+        status = fwk_module_stop_module(module_ctx);
+        if (status != FWK_SUCCESS)
+            return status;
```
tarek-arm, line 510: Brackets, please.

tarek-arm (changes requested):

tarek-arm (changes requested):
in `module/bootloader/sr/mod_bootloader.c` there is a call to `fwk_interrupt_global_disable()`
You need to check the return value to avoid MISRA violation.

vingu-linaro:
> in `module/bootloader/sr/mod_bootloader.c` there is a call to `fwk_interrupt_global_disable()` You need to check the return value to avoid MISRA violation.

I'm doing this in 4 files:
module/bootloader/src/mod_bootloader.c:176
product/synquacer/module/synquacer_rom/src/mod_synquacer_rom.c:44
product/n1sdp/module/n1sdp_rom/src/mod_n1sdp_rom.c:54
product/morello/module/morello_rom/src/mod_morello_rom.c:54

I'm going to use 
(void)fwk_interrupt_global_disable()

tarek-arm:
Jenkins: OK to test

arch/arm/armv7-m/include/arch_helpers.h
```diff
@@ -8,15 +8,29 @@
...
 {
-    __asm volatile("cpsie i" : : : "memory");
+   /* Decrement critical_section_nest_level only if in critical section */
+    if (critical_section_nest_level > 0) {
+        critical_section_nest_level--;
+    }
+
+    /* Enable interrupts globally if now outside critical section */
+    if (critical_section_nest_level == 0) {
+        __enable_irq();
```
tarek-arm, line 32: Sorry, I should have mentioned this. You will also need to `#include <fmw_cmsis.h>` to avoid this error:
```
./../../../../arch/arm/armv7-m/include/arch_helpers.h:32:9: error: implicit declaration of function '__enable_irq' [-Werror=implicit-function-declaration]
   32 |         __enable_irq();
```

tarek-arm (changes requested):

tarek-arm:
Jenkins: OK to test

vingu-linaro:
With https://github.com/ARM-software/SCP-firmware/pull/651 being merged, we can close this one

PR #633, opened 2022-04-05
mod_scmi_clock: Fix Resource Permissions returned error

tgonzalezorlandoarm:
Some scmi tests are failing when building with resource
permissions enabled for some platforms.
When trying to access an incorrect clock or sending an
invalid parameter, the NOT_FOUND or the INVALID_PARAMETER
returns are expected by the test instead of the DENIED error.

Fix the logic so that the first thing that each clock handler does
is check whether the requested clock is available and only then
check whether the correct permissions are given to operate on said
clock.

The disadvantage of this implementation is that every new handler
will need to make sure that permissions are satisfied, while
currently we only perform the permissions check before moving to
the respective handlers. However, placing this function call in the
message_handler would imply having to get the clock_id and check
whether it exists/is valid, duplicating
code in the case where resource permissions are enabled.
Hence, this is the preferred solution.

Signed-off-by: Tomás Agustín González Orlando <tomasagustin.gonzalezorlando@arm.com>
Change-Id: I48dda155e64108555ec933451ea7b888ca64e367

### DIFF
--- a/module/scmi_clock/src/mod_scmi_clock.c
+++ b/module/scmi_clock/src/mod_scmi_clock.c
@@ -619,48 +619,13 @@ FWK_WEAK int mod_scmi_clock_config_set_policy(
 
 #ifdef BUILD_HAS_MOD_RESOURCE_PERMS
 
-/*
- * SCMI Resource Permissions handler
- */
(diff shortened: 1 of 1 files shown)


nicola-mazzucato-arm:
Jenkins: ok to test

tarek-arm (approved):

PR #639, opened 2022-04-17
cli: stop alarm when cli is open

yaronm-hailo:
The debugger cli uses an alarm that periodically read a character from UART
to check if ctrl+E was pressed. However, the alarm is still being called
when the CLI is opened - which causes the CLI to occasionally lose characters.

### DIFF
--- a/module/debugger_cli/src/mod_debugger_cli.c
+++ b/module/debugger_cli/src/mod_debugger_cli.c
@@ -53,6 +53,30 @@ static void alarm_callback(uintptr_t module_idx)
     }
 }
 
+static int start_alarm(fwk_id_t id)
+{
+    const struct mod_debugger_cli_module_config *module_config;
+
+    /* Retrieve the module config as specified by the platform config.c file */
+    module_config = fwk_module_get_data(id);
+
+    /* Start the UART polling alarm, recurring with the given time period */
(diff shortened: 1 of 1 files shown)


tarek-arm:
Jenkins: OK to test

module/debugger_cli/src/mod_debugger_cli.c
```diff
@@ -103,7 +119,9 @@ static int debugger_cli_process_event(const struct fwk_event *event,
     case DEBUGGER_CLI_INTERNAL_EVENT_IDX_ENTER_DEBUGGER:
         /* Start the CLI, blocking the rest of the event queue */
         cli_print("[CLI_DEBUGGER_MODULE] Entering CLI\n");
+        stop_alarm(event->target_id);
```
nicola-mazzucato-arm, line 122: Please make sure you use the returned value from start_ & stop_ functions.
Thanks for the improvements :)

nicola-mazzucato-arm (changes requested):

module/debugger_cli/src/mod_debugger_cli.c
yaronm-hailo: I pushed a new version with checks for the return values of the start & stop functions

tarek-arm:
Jenkins: OK to test

tarek-arm (approved):

PR #645, opened 2022-04-25
morello: fix timestamp in debug logs

manojkumar-arm:
The generic timer instance is not hooked to the framework time driver
which makes the timestamp to be printed as zeros all the time.

This patch fixes the gap to let framework print correct timestamps
during debug logs.

Signed-off-by: Manoj Kumar <manoj.kumar3@arm.com>
Change-Id: Iacab58a01999be8f71358d13e48c1517dba02b64


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #651, opened 2022-05-19
Add scp subsystem option and module stop function 

tarek-arm:
This is an updated version of PR #631 after fixing static analysis issues.

@vingu-linaro can you please test this PR one more time to make sure we haven't broken anything?


vingu-linaro:
I have been tested the PR and It works with my different setups

girishpathak (approved):

PR #657, opened 2022-06-20
RFC v2 - using devicetree to configure SCP firmware

mikel-armbb:
These patches contain an update to the firmware cmake build system that integrates a step to process device tree files in order to populate config arrays.

The library for reading .dts files has been copied from the Zephyr project, as have some of the macro generation headers.
The processing script dt_gen_config.py is based largely on the similar script in the Zephyr project, with changes to work with the SCP frameworks.

There are now multiple modules that have been converted to use device tree configuration,.
These are  used in the product/juno/scp_ramfw.

There are two new cmake list variables used to select modules to be configured by .dts in the product Firmware.cmake :-
SCP_DT_CONFIG_MODULES_STD - for statically configuration tables.
SCP_DT_CONFIG_MODULES_DYN - for dynamically configuration tables.

When a module is selected in one of these lists the relevant config_xxx.c (e.g. config_pl011.c) file needs to be dropped from the CMakelists file for the product.

These lists are then processed by the common framework build files to call the .dts operations.

For any converted module, once selected for .dts support, an additional config_dt_<module>.c is built alongside the main <module> common code. This replaces the data tables from the original files that are linked via the automatic generation in cmake.

The data tables are built from macros generated by the ,dts interpretation code.

The clock module in the Juno scp_ramfw product uses a dynamic table setup to make runtime changes.

The main Zephyr devicetree library has been moved to the /contrib part of the file structure, and the build process updated to correctly use it


tarek-arm:
There will be an updated PR. Closing this one.

PR #667, opened 2022-06-27
 unit_test: integrate unity testing framework 

brett-warren-arm:
New dependencies are added via contrib/cmock:

    - CMock, a mock/stub generator and runtime, designed for use with:
    - Unity Test, a simple unit test framework built for C.

These are integrated into the project via unit_test, which provides:
    - configuration for the framework, including project-wide mocks
    - tools for generating new mocks
    - cmake build system configuration allowing modules to be added to the
      test framework
    - doccumentation instructing how to configure and run the test suite and
      add new module test directories.

The new test framework can be built and run via the 'mod_test' target,
and it is also run in conjuction with framework tests under the existing
'test' target (to build framework tests independently, use 'fwk_test' target).


nicola-mazzucato-arm:
@brett-warren-arm you may need to rebase on latest and re-push the PR for the CI to pass, thanks.

PR #683, opened 2022-07-22
Fix number of PCIe enabled IO-Macros for RD-N2-Cfg1

vivgau01:
On RD-N2-Cfg1 (variant 1), non-pcie devices are behind the IO Macro #1. The PCIe devices are only behind IO Macro#0.
So remove initialization of PCIe configurations for IO Macro #1.


tarek-arm:
Jenkins: OK to test

nicola-mazzucato-arm (approved):

PR #686, opened 2022-07-25
cmn700: fix SYS_CACHE_GRP_SN_NODEID calculation

tonykn-arm:
Every internal RN-SAM node has 32 SYS_CACHE_GRP_SN_NODEID registers to
store corresponding SNF ids. Each SYS_CACHE_GRP_SN_NODEID register
targets four SN node IDs. Previously, the value for each register was
calculated iteratively and was stored in the sn_nodeid_group[group]
array temporarily. Once the entire calculation was over, the contents of
the array were assigned to the corresponding SYS_CACHE_GRP_SN_NODEID
registers.

The values in the sn_nodeid_group[group] array were not cleared before
the calculation of the next RN-SAM nodes. This caused incorrect
calculation of the SYS_CACHE_GRP_SN_NODEID register values for the next
RN-SAM nodes.

To fix this, this patch directly calculates SYS_CACHE_GRP_SN_NODEID
values for each cache group from the values in the snf_table array. The
array sn_nodeid group is removed from the device context since it is no
longer used for SYS_CACHE_GRP_SN_NODEID calculation.

Change-Id: Ic9dc4b4520c09aad4589576b9eeb644e68e8afb3
Signed-off-by: Tony K Nadackal <tony.nadackal@arm.com>

### DIFF
--- a/module/cmn700/include/internal/cmn700_ctx.h
+++ b/module/cmn700/include/internal/cmn700_ctx.h
@@ -1,6 +1,6 @@
 /*
  * Arm SCP/MCP Software
- * Copyright (c) 2020-2021, Arm Limited and Contributors. All rights reserved.
+ * Copyright (c) 2020-2022, Arm Limited and Contributors. All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  *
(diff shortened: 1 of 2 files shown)


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #688, opened 2022-07-29
Master sp805 wdt

tonykn-arm:


tonykn-arm:
The changes are already merged in another pull request

PR #694, opened 2022-08-09
product/tc2: Add support for TC2/RSS boot flow

tintuthomas-arm:
Add TC2 BL1 module that implements the boot flow, with MHU (doorbell)
signalling to/from RSS.

BL1 will signal the RSS to indicate that startup is complete, then wait
for another signal from RSS before powering on the AP CPUs.

Signed-off-by: Anders Dellien <anders.dellien@arm.com>
Change-Id: I2747bd0d07e867764c9a498ae87a985f4b6b7035
Signed-off-by: Tintu Thomas <tintu.thomas@arm.com>


girishpathak (approved):

girishpathak:
jenkins: ok to test

PR #695, opened 2022-08-12
Fast Channels support

tgonzalezorlandoarm:
This pull request:

1) adds support for the Fast Channel Extension to the MHUv3 drivers.
2) Fixes an issue with the fast channel defines in the transport layer.
3) Fixes an issue with the timer base address of the LCP.


framework/CMakeLists.txt
```diff
@@ -59,7 +59,12 @@ if(SCP_ENABLE_SCMI_NOTIFICATIONS)
     endif()
 endif()
 
+if(SCP_ENABLE_FAST_CHANNELS_TRANSPORT)
```
nicola-mazzucato-arm, line 62: I think we need a different name for this.

When building a platform, we may want to add support for features, and the feature here is FastChannels as the SCMI spec describes. A platform may want to support SCMI Performance FastChannels and therefore they would set -DSCP_ENABLE_SCMI_PERF_FAST_CHANNELS=yes.
This would instruct the build system to enable the protocol-specific FastChannel support. But because also the transport module offers an abstraction of FastChannels, those relevant sections would need to be enabled - and this is what happens here. The issue is that I don't see a case where we would enable FastChannels support in transport module w/o having enabled the relevant sections for the specific protocol. Based on this, IMO we should rename SCP_ENABLE_FAST_CHANNELS_TRANSPORT to SCP_ENABLE_FAST_CHANNELS as a generic option that can be used by all the modules that offer functionalities for FastChannels regardless the chosen protocol.
Also, like you have correctly done, SCP_ENABLE_FAST_CHANNELS would be exclusively enabled by the build system once we choose to build with SCP_ENABLE_SCMI_PERF_FAST_CHANNELS or any other SCMI protocol that the spec describes.
To summarize, the protocol-specific build option should enable a generic build flag that modules can consume if they provide support for FastChannels in a generic manner.

Hope it makes sense.

nicola-mazzucato-arm (changes requested):

framework/CMakeLists.txt
tgonzalezorlandoarm: Makes sense, will update accordingly

nicola-mazzucato-arm (changes requested):
The patch "module/transport: fix fast channel compilation setup" should be renamed to something like "scp: Fix fast channel compilation setup" or similar. It does not directly affect the transport module, but only CMake files.
Also worth adding some infor about this build option in doc/readme.md along the other ones.
Thanks in advance.

nicola-mazzucato-arm (approved):
Thanks for the changes. LGTM

PR #713, opened 2022-09-26
scmi_pd: Fix handling of device power OFF

jpbrucker:
Since commit 656f089b97d5 ("scmi_pd:Add power domain set state sync request handling"), the scmi_power_domain module incorrectly masks the power state parameter for device power domains.

The SCMI specification defines a parameter of value 0 as the ON state, and value 0x40000000 as the OFF state. Any other value for bits 27:0 is IMPLEMENTATION DEFINED. Currently, in order to convert the SCMI parameter into an internal state, scmi_pd_power_state_set_policy() masks all bits but 27:0, so that a OFF parameter incorrectly matches MOD_PD_STATE_ON. To fix it, compare the complete parameter value against pd_state_to_scmi_dev_state.

Signed-off-by: Jean-Philippe Brucker <jean-philippe@linaro.org>


tarek-arm:
Jenkins: OK to test

tarek-arm:
Hi @jpbrucker, thanks for finding this error. We can merge the patch after we finish the release in a week or so.

tarek-arm (approved):

PR #721, opened 2022-11-07
Use add_subdirectory to include individual test modules and other scope related fix

girishpathak:
This PR changes unit_test/CMakeLists.txt to include individual test modules using CMake's 'add_subdirectory' command instead of 'include' to keep the scope of the test module from the parent. It also fixes minor FWK_SRC scoping, see commit [module_common: fix FWK_SRC scoping](https://github.com/ARM-software/SCP-firmware/commit/f97acf5b559df88cfdb370ab2b316966468c6dbd) for details


tarek-arm:
Jenkins: OK to test

girishpathak:
jenkins: ok to test

tarek-arm (approved):

PR #723, opened 2022-11-16
scmi_clock: Update comments about source of event

BY1994:
Hi, I want to change the comments in scmi_clock slightly. In current status, event source is missing for the first comment.
When I looked module/scmi_perf/src/mod_scmi_perf.c file (2304 - 2310 lines), the comments helped me a lot to understand the codes. So I want to suggest changing the comments as the same format.
Any feedback is welcome including suggestion of commit message.


nicola-mazzucato-arm (commented):
Hi @BY1994 ,
thanks for your PR.
I can see this is a trivial shift of comments by a few lines without adding clarifications.
Is there anything else it's not clear that needs explanation?
thanks

BY1994:
Hi @nicola-mazzucato-arm, thank you for your feedback! I agree that just changing comments is trivial. I will close this request. Thanks!!

PR #732, opened 2022-12-05
fixes missing-prototypes reports

etienne-lms:
Few changes in framework and some modules to address build issues reported when building ith OP-TEE, its build environment and toolchain. I have not covered all modules, only those currently embedded in the optee-* products.

### DIFF
--- a/module/scmi_system_power_req/test/mod_scmi_system_power_req_extra.h
+++ b/module/scmi_system_power_req/test/mod_scmi_system_power_req_extra.h
@@ -9,19 +9,6 @@
  */
 #include <mod_scmi_system_power_req.h>
 
-/*! Send an SCMI message */
-int scmi_send_message(
-    uint8_t scmi_message_id,
-    uint8_t scmi_protocol_id,
-    uint8_t token,
-    fwk_id_t service_id,
-    const void *payload,
-    size_t payload_size,
-    bool request_ack_by_interrupt);
-
-/*! Handle response SCMI message */
-int response_message_handler(fwk_id_t service_id);
-
 /*
  * used as a mock for message handlers
  */
--- a/framework/src/fwk_delayed_resp.c
+++ b/framework/src/fwk_delayed_resp.c
@@ -5,8 +5,10 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+#include <internal/fwk_delayed_resp.h>
 #include <internal/fwk_module.h>
 
+#include <fwk_core.h>
 #include <fwk_event.h>
 #include <fwk_id.h>
 #include <fwk_interrupt.h>
--- a/framework/src/fwk_string.c
+++ b/framework/src/fwk_string.c
@@ -10,6 +10,7 @@
 
 #include <fwk_assert.h>
 #include <fwk_mm.h>
+#include <fwk_string.h>
 
 #include <stdlib.h>
 
--- a/framework/src/fwk_time.c
+++ b/framework/src/fwk_time.c
@@ -17,7 +17,7 @@ struct {
     const void *driver_ctx; /* Time driver context */
 } fwk_time_ctx;
 
-FWK_CONSTRUCTOR void fwk_time_init(void)
+static FWK_CONSTRUCTOR void fwk_time_init(void)
 {
     struct fwk_time_driver driver = fmw_time_driver(&fwk_time_ctx.driver_ctx);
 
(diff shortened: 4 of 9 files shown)


etienne-lms:
Any update on these changes?

etienne-lms:
I would think that it is better for scp-firmware to declare exported prototypes and more generally follow the toolchains default build warnings recommendation (which evolves with toolchains), unless a good reason prevents that a warning to be addressed.

module/scmi_system_power_req/test/mod_scmi_system_power_req_extra.h
```diff
@@ -9,19 +9,6 @@
  */
 #include <mod_scmi_system_power_req.h>
 
-/*! Send an SCMI message */
```
nicola-mazzucato-arm, line 12: these need to stay as they're part of the tests

nicola-mazzucato-arm (changes requested):
Can you please share which compiler you are using and is it used only locally or is there automation in place?
Thanks

etienne-lms (commented):
Thanks for the feedback. I'll update this P-R.

module/scmi_system_power_req/test/mod_scmi_system_power_req_extra.h
etienne-lms: Discard my comment. This is a test related header file, so fine. I'll discard this change.

etienne-lms: Hmm.. I reconsider my position. 

These 2 prototypes are declared from 2 different header files (here and _mod_scmi_sensor_req_extra.h_). I don't think it is nice. There should be 1 declaration for all sources.
Couldn't _mod_scmi_system_power_req_extra.h_ include _mod_scmi.h_ to get the test setup happy?

etienne-lms (commented):
> Can you please share which compiler you are using and is it used only locally or is there automation in place?

As for automation, OP-TEE CI currently only test builds with SCP-firmware/SCMI stack. There is no effective test execution yet. SCP-fmw/OP-TEE is built for 3 platform flavors. A [Qemu/armv8-A](https://github.com/OP-TEE/optee_os/blob/master/.github/workflows/ci.yml#L130) platform is built in Aarch64 and is quite used in OP-TEE for development/tests/demo sharing. OP-TEE CI runs regression tests with a Qemu/armv8 image but these tests currently don't use/cover SCMI resources. CI also builds SCP-fmw for a [FVP armv8-A](https://github.com/OP-TEE/optee_os/blob/master/.github/workflows/ci.yml#L139) platform and an [STM32MP15 armv7-A](https://github.com/OP-TEE/optee_os/blob/master/.github/workflows/ci.yml#L130) platform both in 32bit. 

If i'm right, CI build environment runs in a virtualized x86/ubuntu container. On x86 build hosts OP-TEE currently uses Arm's GCC 10.2 release, for 32 and 64bit ([_toolchain.mk:46_](https://github.com/OP-TEE/build/blob/3.20.0/toolchain.mk#L46-L55)).
OP-TEE distribs can also build on Aarch64 host machines. In that case we use Arm GCC 10.2 cross toolchain for arm 32bit ([toolchain.mk:86](https://github.com/OP-TEE/build/blob/master/toolchain.mk#L86-L89)) but a Buildroot based toolchain for native Aarch64 ([toolchain.mk:91](https://github.com/OP-TEE/build/blob/master/toolchain.mk#L91-L97)).  There are (crazy?) people building with clang/llvm and one could select it's own toolchain (assuming it's reliable) to build OP-TEE. OP-TEE may one day dump to GCC 11 or 12, or support a range of GCC's and Clang's.

(@jforissier, correct me if i'm wrong)

etienne-lms:
I think all review comments have been addressed.

module/scmi_system_power_req/test/mod_scmi_system_power_req_extra.h
girishpathak: @etienne-lms : can we move to a separate header mod_scmi_ut.h ? and keep this file under module/scmi/test/includes ? 
scmi_sensor and system power test modules CMakeLists.txt then can include a path for this header. This way it won't affect mod_scmi.h

girishpathak (approved):

girishpathak:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Agreed with @etienne-lms to close this PR. The same request will be then opened in gitlab soon.
We will provide more information about what failed for unit testing.
Thanks

PR #733, opened 2022-12-05
arch: optee: fw_assert.h wrapper for fwk_expect() return value

etienne-lms:
Implements wrapper header file fwk_assert.h for optee architecture to overcome the issue building with OP-TEE. OP-TEE OS assert() function does not provide a return value resulting a compilation failure as reported in the below build traces:

/tmp/optee-3.19/build/../scp-firmware/framework/src/fwk_arch.c: In function ‘fwk_arch_init’: lib/libutils/isoc/include/assert.h:20:2: error: expected expression before ‘do’
   20 |  do { \
      |  ^~
/tmp/optee-3.19/build/../scp-firmware/framework/include/fwk_assert.h:147:35: note: in expansion of macro ‘assert’
  147 | #    define fwk_assert(condition) assert(condition)
      |                                   ^~~~~~
/tmp/optee-3.19/build/../scp-firmware/framework/include/fwk_assert.h:180:34: note: in expansion of macro ‘fwk_assert’
  180 | #    define fwk_check(condition) fwk_assert(condition)
      |                                  ^~~~~~~~~~
/tmp/optee-3.19/build/../scp-firmware/framework/include/fwk_assert.h:212:42: note: in expansion of macro ‘fwk_check’
  212 | #    define fwk_expect(condition) (bool)(fwk_check(condition), 1)
      |                                          ^~~~~~~~~
/tmp/optee-3.19/build/../scp-firmware/framework/src/fwk_arch.c:73:10: note: in expansion of macro ‘fwk_expect’
   73 |     if (!fwk_expect(status == FWK_SUCCESS)) {
      |          ^~~~~~~~~~

This change works around the issue by implementing a functional fwk_expect() function using OP-TEE OS standard lib libutils resources.

### DIFF
--- a/arch/none/optee/include/fwk_assert.h
+++ b/arch/none/optee/include/fwk_assert.h
@@ -1,0 +1,40 @@
+/*
+ * Arm SCP/MCP Software
+ * Copyright (c) 2022, Linaro Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef OPTEE_FWK_ASSERT_H
+#define OPTEE_FWK_ASSERT_H
+
+#include <assert.h>
+#include_next <fwk_assert.h>
+
+#if !defined(NDEBUG) && !defined(BUILD_TESTS)
+/*
+ * Redefine fwk_expect() as SCP-firmware implementation expects
+ * assert() macro to return a value fwk_expect() can propgate.
+ *
+ * SCP-firmware documentation regarding fwk_expect():
+ * In debug builds, the macro will evaluate the condition and trigger
+ * an assertion if it fails. In release builds, or if running tests, the
+ * macro will evaluate the condition and discard its result.
+ */
+#undef fwk_expect
+static inline bool _fwk_expect(bool cond, const char *expr,
+			       const char *file, const int line,
+			       const char *func)
+{
+	if (!cond) {
+		_assert_log(expr, file, line, func);
+		_assert_break();
+	}
+
+	return true;
+}
+
+#define fwk_expect(condition) _fwk_expect(condition, #condition, \
+					  __FILE__, __LINE__, __func__)
+#endif /* !NDEBUG && !BUILD_TESTS */
+#endif /* OPTEE_FWK_ASSERT_H */


arch/none/optee/include/fwk_assert.h
```diff
@@ -0,0 +1,39 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Linaro Limited
+ *
```
etienne-lms, line 4: i will move this empty line

arch/none/optee/include/fwk_assert.h
```diff
@@ -0,0 +1,39 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
etienne-lms, line 1: I'll switch to BSD-3-Clause

arch/none/optee/include/fwk_assert.h
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+#ifndef OPTEE_FWK_ASSERT_H
+#define OPTEE_FWK_ASSERT_H
+
+#include <assert.h>
+#include_next <fwk_assert.h>
+
+#ifndef NDEBUG
```
etienne-lms, line 13: should be `#if defined(NDEBUG) || defined(BUILD_TESTS)`

etienne-lms:
comments addressed.

etienne-lms:
Is this change acceptable in SCP-firmware?

girishpathak:
@etienne-lms I tried to build "SCMI server from SCP-firmware (take 2) #5686" and effectively disabled the workaround
however, I was able to make debug/release optee builds without any errors

here is my build command

` make PLATFORM=stm32mp1-157C_DK2 CFG_SCP_FIRMWARE=y CFG_SCMI_SCPFW=y optee-os DEBUG=1`

Below is fwk_assert.h `optee_os/core/lib/scmi-server/include/fwk_assert.h` in my setup

`
#ifndef SCMI_SERVER_FWK_ASSERT_H
#define SCMI_SERVER_FWK_ASSERT_H

#include <assert.h>
#include_next <fwk_assert.h>

#endif /* SCMI_SERVER_FWK_ASSERT_H */
~                                             
`

what is the command to reproduce the build error ?

etienne-lms:
Hello @girishpathak, 
The initial build error would occur if neither SCP-firmware nor optee_os source trees provide such a wraper header file. For example if you remove both _scp-firmware/arch/none/optee/include/fwk_assert.h_ and _optee_os/core/lib/scmi-server/include/fwk_assert.h_ and build with debug with (`CFG_TEE_CORE_DEBUG=y`):
```
$ rm ../optee_os/core/lib/scmi-server/include/fwk_assert.h ../scp-firmware/arch/none/optee/include/fwk_assert.h
$ make PLATFORM=stm32mp1-157C_DK2 CFG_SCMI_SCPFW=y CFG_TEE_CORE_DEBUG=y optee-os
...
lib/libutils/isoc/include/assert.h:20:2: error: expected expression before ‘do’
   20 |  do { \
      |  ^~
```
The question is where to host that file? Optee_os.git or SCP-firmware.git? I initially hosted it in optee_os, with P-R #5686. But files named _fwk_xxx.h_  rather belongs to scp-firmware, this is why I created this P-R #733 providing _fwk_assert.h_ from scp-fmw **arch/optee/include/** directory. Not a product include directory but the optee archi include dir.

Regarding OP-TEE, P-R #5686 integrates CI build test means on optee + scp-firmware for _product/optee-fvp/_ and _product/optee-stm32mp1/_. The OP-TEE P-R still includes an instance of _fwk_assert.h_ for CI test to run successfully using SCP-firmware latest master. The plan is to remove that _optee_os/.../fwk_assert.h_ file and rely on the one from scp-firmware/arch/optee if/when current P-R i accepted and merged.

girishpathak:
> make PLATFORM=stm32mp1-157C_DK2 CFG_SCMI_SCPFW=y CFG_TEE_CORE_DEBUG=y optee-os

Hi @etienne-lms I've looked into this again in detail,  the problem is  in `./lib/libutils/isoc/include/assert.h`
the implementation of the macro `assert` defers from toolchain implementation, it does not consider macro `assert` to be used in 'if()` and I think that is the problem, as far as SCP firmware is concerned the implementation of `fwk_assert.h` should work with GCC toolchain supplied standard `assert.h`. To be honest `./lib/libutils/isoc/include/assert.h` should be changed to behave as the toolchain provided `assert.h`,  so in IMO if you need to have a workaround then it must be in optee tree somewhere. The workaround suggested in this PR(733) hides the real issue.

etienne-lms:
Hi @girishpathak. Thanks for the feedback. I understand your view. I will discuss this with OP-TEE maintainers and keep the fw_assert.h wrapper in optee_os source tree until we find a cleaner solution.

PR #737, opened 2022-12-08
module: cmake: add option to replace modules

Joe-Zhucg:
Some platforms have special feature need to slightly modify upstream module. Add option to replace with the module defined in product path.

Change-Id: I1dc89ca49eecb1f2fb2e2537234cea7f98116c1c
Signed-off-by: Joe Zhu <Chunguang.Zhu@verisilicon.com>

### DIFF
--- a/module/CMakeLists.txt
+++ b/module/CMakeLists.txt
@@ -86,6 +86,14 @@ list(APPEND SCP_MODULE_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/traffic_cop")
 list(APPEND SCP_MODULE_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/transport")
 list(APPEND SCP_MODULE_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/voltage_domain")
 
+#
+# Remove paths if this variable is defined by the platform target's
+# Firmware.cmake
+#
+if(SCP_MODULE_EXCLUDE_PATHS)
+    list(REMOVE_ITEM SCP_MODULE_PATHS ${SCP_MODULE_EXCLUDE_PATHS})
+endif()
+
 foreach(source_dir IN LISTS SCP_MODULE_PATHS)
     unset(SCP_MODULE)
 


module/CMakeLists.txt
```diff
@@ -86,6 +86,10 @@ list(APPEND SCP_MODULE_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/traffic_cop")
 list(APPEND SCP_MODULE_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/transport")
 list(APPEND SCP_MODULE_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/voltage_domain")
 
+if (SCP_MODULE_EXCLUDE_PATHS)
```
girishpathak, line 89: IIUC, SCP_MODULE_EXCLUDE_PATHS will be defined in Firmware.cmake ? in that case can you add some comment
e.g.

```
# Remove paths if this variable is defined by the platform target's Firmware.cmake
if (SCP_MODULE_EXCLUDE_PATHS)
```

girishpathak (changes requested):

module/CMakeLists.txt
Joe-Zhucg: Correct, SCP_MODULE_EXCLUDE_PATHS is defined in Firmware.cmake. Will update patch to add comment.

girishpathak (approved):

girishpathak:
Jenkins: ok to test

nicola-mazzucato-arm:
Hi, I would be happy to merge this PR. I am just wondering if we should provide some description of this newly added option in doc/cmake_readme.md?

Joe-Zhucg:
How about add below in "Modifying build options" section?

- Modify default module implementation for special requirement.
>
> Some platforms need to slightly modify default module to support new features.
> It is preferred to replace with modified module under product path instead of changing the
> code under module/ directly. SCP_MODULE_EXCLUDE_PATHS provides one option to remove 
> default module path.
> 
> e.g. add below in Firmware.cmake to replace timer module
> ```
> list(APPEND SCP_MODULE_EXCLUDE_PATHS "${CMAKE_SOURCE_DIR}/module/timer")
> 
> list(PREPEND SCP_MODULE_PATHS "${CMAKE_CURRENT_LIST_DIR}/../module/timer")
> ```

PR #742, opened 2022-12-26
product/optee-fvp: fix scmi clock init state

vingu-linaro:
Align SCMI clock init value with MOCK clock and system PLL which are enabled by default.

SCMI compliance test fails without this change.

### DIFF
--- a/product/optee-fvp/fw/config_scmi_clock.c
+++ b/product/optee-fvp/fw/config_scmi_clock.c
@@ -1,6 +1,6 @@
 /*
  * Arm SCP/MCP Software
- * Copyright (c) 2022, Linaro Limited and Contributors. All rights reserved.
+ * Copyright (c) 2022-2023, Linaro Limited and Contributors. All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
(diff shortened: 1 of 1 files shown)


vingu-linaro:
@nicola-mazzucato-arm This one is quite similar to https://github.com/ARM-software/SCP-firmware/pull/743 but for the clock

nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

vingu-linaro:
while testing per channel scmi raw mode, I noticed that I have fixed one channel (the one use by default) but not the other (normally only used for performance) ...
Let me push a new version that will fix that for both channel

nicola-mazzucato-arm:
Glad that the scmi raw mode is coming useful :)
No worries, I'll retrigger when ready.
Thanks

vingu-linaro:
I have to keep the current version of the patch: When I try to fix the other channel, both fail to pass the scmi test. This is linked to how the scmi clock and their init states are managed in scmi_clock module. I'm going to create a issue to describe my problem

nicola-mazzucato-arm:
Jenkins: ok to test

PR #744, opened 2022-12-28
module/scmi_sensor: reserved field must be null

vingu-linaro:
Reserved field should be considered null. SCMI compliance test reported non conformant command if reserved fields are not clear but SCMI server return success.

Unlike other reserved fields, it's not written that this one must be zeroed. I wonder if this is just a miss in the spec as most of them  must be zeroed. 

### DIFF
--- a/module/scmi_sensor/src/mod_scmi_sensor.c
+++ b/module/scmi_sensor/src/mod_scmi_sensor.c
@@ -1,6 +1,6 @@
 /*
  * Arm SCP/MCP Software
- * Copyright (c) 2015-2022, Arm Limited and Contributors. All rights reserved.
+ * Copyright (c) 2015-2023, Arm Limited and Contributors. All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  *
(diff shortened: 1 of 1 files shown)


nicola-mazzucato-arm:
I am not sure about this. The SCMI server does not have an obligation to verify that reserved fields must be 0. It's rather the client that needs to fulfill the expectation - as described in the spec - that reserved fields are set to 0.
Or am I missing something?

vingu-linaro:
> I am not sure about this. The SCMI server does not have an obligation to verify that reserved fields must be 0. It's rather the client that needs to fulfill the expectation - as described in the spec - that reserved fields are set to 0. Or am I missing something?

I don't have a strong opinion but I can imagine that the goal is to make the scmi server as robust as possible. If a reserved bit is  set, it could be that the client expects to discuss with a more recent version of the scmi spec than the one implemented in the server and the client can expect some behavior that the scmi server will not be able to support whereas it has returned success. Or at the opposite, once you will upgrade the scmi server with a spec version that makes use of this reserved bit, the behavior can change and make the client starts to fail.

vingu-linaro:
after offlist discussion, it has been concluded that the scmi compliance test suite will be updated to not raised non conformant test for this case

PR #762, opened 2023-02-16
module: timer: re-enable timer in isr if no alarm

Joe-Zhucg:
32 bits timer uses interrupt to support 64 bits counter

### DIFF
--- a/module/timer/src/mod_timer.c
+++ b/module/timer/src/mod_timer.c
@@ -557,8 +557,14 @@ static void timer_isr(uintptr_t ctx_ptr)
         (struct alarm_sub_element_ctx *)fwk_list_pop_head(&ctx->alarms_active);
 
     if (alarm == NULL) {
-        /* Timer interrupt triggered without any alarm in the active queue */
-        fwk_unexpected();
+        /*
+         * 32 bits timer need interrupt to maintain 64 bits counter.
+         * just re-enable timer if no alarm in the active queue
+         */
+        status = ctx->driver->enable(ctx->driver_dev_id);
+        if (status != FWK_SUCCESS) {
+            FWK_LOG_DEBUG("[Timer] %s @%d", __func__, __LINE__);
+        }
         return;
     }
 


nicola-mazzucato-arm (commented):
Thanks @Joe-Zhucg for this proposed change. I would like to understand a bit more about the use-case. IIUC you start the timer in a module but you don't have a callback associated with it? The current implementation assumes that you should have a callback assigned when you start a timer, otherwise you would hit the unexpected stopper. How did you start the timer without a callback available? Could you not have a dummy callback that does nothing in your implementation? Any details to understand better your use case would be very helpful.
Thanks!

Joe-Zhucg:
Alarm API has callback to do something after alarm timeout. I think alarm API is used by timer client. For HW timer driver, it needs to implement timer driver API. 
We have one 32 bits HW timer which count from load value (max is 0xFFFFFFFF) to 0 and generate an interrupt when counter is 0. If the timer cycle is 1 uS, it will be timeout quickly. So I use one 64 bits SW counter in module context to extend the timer. This SW counter will be calculated after HW timer timeout.  HW timer is enabled in HW timer module start function. If there may be no client to enable alarm, HW timer driver has no chance to calculate the SW counter.

nicola-mazzucato-arm:
@Joe-Zhucg IIUC the 64-bit timer is not in the common implementation. Is it something that you added for the above purpose? If so, then the best way to handle your case is to add an alarm that does not do anything, then the rest of the ISR will re-enable the alarm.
Adding a sub-element in the timer configuration (sub_element_count) and registering that timer at start (alarm_api->start) should give you the ability to have a custom ISR (alarm_callback(uintptr_t param)) that simply won't do anything byut will preserve the timer functionality (along with the enable that you need). You may also want to think if in that callback you prefer to handle/inc the 64b timer you are using.
Hope it helps.
Thanks

Joe-Zhucg:
Thanks a lot!
32 bits timer will be timeout soon. So it can not be used for system timer.
It looks OK to add additional alarm to handle timer counter calculation. But I hope the timer driver can be simple.
BTW, why handle timer interrupt in HAL?  If timer driver can handle interrupt, it can do any special process to cover different hardware feature. Then it can call one HAL API in interrupt handler to trigger HAL state change.


Best Regards,
Joe Zhu

发件人: Nicola Mazzucato ***@***.***
发送时间: 2023年4月20日 19:46
收件人: ARM-software/SCP-firmware
抄送: Zhu, Joe; Mention
主题: Re: [ARM-software/SCP-firmware] module: timer: re-enable timer in isr if no alarm (PR #762)


@Joe-Zhucg<https://github.com/Joe-Zhucg> IIUC the 64-bit timer is not in the common implementation. Is it something that you added for the above purpose? If so, then the best way to handle your case is to add an alarm that does not do anything, then the rest of the ISR will re-enable the alarm.
Adding a sub-element in the timer configuration (sub_element_count) and registering that timer at start (alarm_api->start) should give you the ability to have a custom ISR (alarm_callback(uintptr_t param)) that simply won't do anything byut will preserve the timer functionality (along with the enable that you need). You may also want to think if in that callback you prefer to handle/inc the 64b timer you are using.
Hope it helps.
Thanks

PR #765, opened 2023-02-23
tc2: dvfs and hunter elp

benhor01:
Add support for hunter elp core in total compute 2 (tc2) as this is now used in place of hunter for core 7 and add dvfs power numbers for all tc2 cpu cores.


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #768, opened 2023-03-09
scmi_clock: Fix policy reference counter mismatch

leandro-arm:
The reference count for each clock is stored into an internal array, clock_count[], whose size is "number of available clocks" (not SCMI).

In that table, we are mistakenly using an index that is taken from the agent's number of clocks. That leads to a mismatch of clock reference count in some cases.

Fix this by using the clock index that corresponds to the agent's device index to determine to correct location in clock_count[] table to update the reference count.

At the same time, clean-up some of the variables' name to ease the readability of the various clocks.

Change-Id: I9819891fe2237abb252764dadd4a1d36302bc4bf

### DIFF
--- a/module/scmi_clock/src/mod_scmi_clock.c
+++ b/module/scmi_clock/src/mod_scmi_clock.c
@@ -46,9 +46,9 @@ struct clock_operations {
     enum mod_clock_state state;
 
     /*
-     * The clock device for this operation.
+     * The SCMI clock index for this operation.
      */
-    uint32_t clock_dev_id;
+    uint32_t scmi_clock_idx;
 
     /*
      * Request type for this operation.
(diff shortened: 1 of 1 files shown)


vingu-linaro:
@nicola-mazzucato-arm and @leandro-arm 
Thanks, I 'm going to run some tests and will come back to you

module/scmi_clock/src/mod_scmi_clock.c
```diff
@@ -610,11 +642,11 @@ FWK_WEAK int mod_scmi_clock_config_set_policy(
...
          * This is the last agent to set the clock STOPPED.
          */
-        if (clock_count[clock_dev_id] != 1) {
+        if (clock_count[clock_idx] != 1) {
             status = FWK_E_STATE;
         }
         if (policy_commit == MOD_SCMI_CLOCK_POST_MESSAGE_HANDLER) {
-            clock_count[clock_dev_id]--;
+            clock_count[clock_idx]--;
         }
```
vingu-linaro, line 650: The decrement should be done only in post message handler otherwise you will never be able to decrement clock_count[clock_idx] because if clock_count[clock_idx] != 1 you don't do post the message  so once clock_count[clock_idx] > 1 you will never decrement it

tarek-arm (approved):

PR #771, opened 2023-03-20
[RFC] Scp 2.11 dts build rfc 003

mikel-armbb:
RFC  - for device tree config

Updated version of configuration of SCP firmware build using device tree

Contains documentation in device_tree_support.md, alongside doxygen comments in fwk_devicetree.h and fwk_dt_config_common.h. These will all appear when documentation is built.

Multiple modules supplied as worked examples, using the updated build flow - significant changes since last PR.


leandro-arm (changes requested):
Remove this please, CMSIS needs to be 5.8.0

leandro-arm (changes requested):
Is it possible to add python-devicetree as a git submodule?

PR #772, opened 2023-03-22
scmi_clock: clean up policy functions and return the SCMI agent's state instead of the physical state 

mohamedasaker-arm:
This patchset:

- cleans up the policy functions from the `scmi_clock` module. As seen in `c49f043`, due to the code complexity of the function can potentially introduce defects that can be difficult to spot and maintain. Refactoring `mod_scmi_clock_config_set_policy` function also helps to reduce the number of function calls and processing time, which can improve performance and efficiency.
Furthermore, this commit refactors the module by removing all unnecessary parameters in helper functions.
In addition, removing the FWK_WEAK to enforce the current policy until further requirements.
- Return the SCMI agent's state instead of the physical state.
  The current method of reading the clock state calls the Clock HAL directly,  which can cause inconsistencies
  between what is being read by Clock Hal and what is being set in the SCMI agent state table.
  To address this, the existing `scmi_clock_get_agent_clock_state` helper function is used to read the clock 
  state in relation to the requesting SCMI agent.  i.e. function returns the state expected by SCMI agent.

### DIFF
--- a/module/scmi_clock/src/mod_scmi_clock.c
+++ b/module/scmi_clock/src/mod_scmi_clock.c
@@ -1506,6 +1506,8 @@ static int process_request_event(const struct fwk_event *event)
     struct event_set_state_request_data set_state_data;
     fwk_id_t service_id;
     enum scmi_clock_event_idx event_id_type;
+    unsigned int scmi_clock_idx, agent_id;
+    enum mod_clock_state agent_clock_state;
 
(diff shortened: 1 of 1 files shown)


tarek-arm:
Jenkins: OK to test

girishpathak:
jenkins: ok to test

girishpathak (approved):

PR #778, opened 2023-03-29
module/mhu3: fix bind function

girishpathak:
In the mhu3_bind() function, the transport_api pointer variable is passed to the fwk_module_bind() instead of the pointer address. This causes an error when the framework bind function checks if the API pointer is NULL. This patch fixes this error.

Change-Id: I774b0ec5104232918b73602b54ce74a86617b185


nicola-mazzucato-arm:
Hi @girishpathak can you please rebase on latest mainline and repush?
Thanks

nicola-mazzucato-arm:
Jenkins: ok to test

tarek-arm:
@girishpathak we are going to close this

PR #779, opened 2023-04-06
cli: Prepend carriage return to line feed

mohamedasaker-arm:
After c15ff7e9505e3544daa4ac5329f302e536fe7acf
which removes the prepending carriage return from the drivers. It moves the responsibility of prepending carriage return to the user of the driver. This patch is adding this logic to the cli.


girishpathak (approved):

girishpathak:
jenkins: ok to test

PR #794, opened 2023-05-19
module: timer: add API to update timer status without alarm

Joe-Zhucg:
Some device need interrupt to update internal status. Add new driver API to do this if there is no acitve alarm.

Change-Id: I6a3106beb8dce8de7de29dfa5ff5742a9fd9692b


nicola-mazzucato-arm:
Hi @Joe-Zhucg 
if I understand correctly, you need to update the driver's internals every time there's an interrupt.
What I would like to understand better is how you use the timer feature without the alarm.
Can you please provide a bit more info?
How do you set and enable the timer driver if you don't make use of the alarms?

Thanks

Joe-Zhucg:
Correct, I want to use this API to update internal counter. My platform only has one 32 bits HW timer and it will timeout soon.  Driver uses one 64 bits SW counter and this 32 bits timer counter register to simulate 64 bits timer. 64 bits SW counter will be updated in each timeout interrupt even it is not an alarm interrupt.   
Timer starts at initialize function, count from 0 and initial timeout is 0xffffffff.  With 24MHz timer clock input, it will timeout in 3 minutes.

nicola-mazzucato-arm:
Will bring this PR over to gitlab where we will continue the conversation.
In principle LGTM, but I may ask minor renaming.
Thanks @Joe-Zhucg

nicola-mazzucato-arm:
@Joe-Zhucg can you please login into the gitlab SCP-firmware repo so we can continue the conversation there?
Thanks

PR #800, opened 2023-06-09
n1sdp: Remove exception_handler function from pcie_enumeration.c

deepthip13:
The exception_handler function expects the stack to be setup with the core registers in a certain order by the parent caller function (arch_exception_invalid). For release builds, as part of the optimisations, the compiler inlines the exception_handler function within the caller - arch_exception_invalid function. But this is not true of debug builds, and where because of the explicit function call, the lr is pushed onto the stack. This changes the order of registers on the stack and defies what is expected by the callee (exception_handler).

This change gets rids of the exception_handler function, moves the handling within the callee into the caller in order to do away with an explicit function call, and retain the expected order of registers pushed on to the stack.

Change-Id: I72d380af649e9f5e597793a40dc86a2fe2a2f8ae


tarek-arm:
Jenkins: OK to test

tarek-arm (approved):

PR #807, opened 2023-06-21
cmn700: use offset adjusted address for remote chips

nissha03:
Adjust the address of SYSCACHE type region with chip address space offset while mapping the regions in CMN.

This change fixed the bug introduced in the previous change which wrongly maps local system cache address in remote chip's without adjusting it with chip address space.

Change-Id: Ic3b47200b1d1646b8d0e746f0ba55caf7b616c5f


nissha03:
Jenkins: OK to test

nicola-mazzucato-arm:
Jenkins: OK to test

nicola-mazzucato-arm (approved):

PR #810, opened 2023-07-06
CMake: Add target info for fwk_test and mod_test in help

abhishek-arm:
This change adds missing target info for fwk_test and mod_test in cmake help to inform user about these unit testing targets.


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #812, opened 2023-07-11
SCMI-Perf Fastchannels, Ops split & Fixes

nicola-mazzucato-arm:
This PR brings some improvements in code separation for SCMI-Perf and some fixes in perf-plugins-handler.


nicola-mazzucato-arm:
@vingu-linaro Purely for the record, this PR fails the test due to the way that SCP is built for OPTEE.

I suspect that the optee build system does not currently fetch the additional files that one of the patches has added [1], for example scmi_perf_protocol_ops.c. I believe that the optee b/s will need to be slightly improved in order to include "any" files that are found under a directory that the scp-under-optee is concerned to build.

Thanks

[1] scmi_perf: Move protocol ops to separate file: https://github.com/ARM-software/SCP-firmware/pull/812/commits/fc434fd2741e2f771428b06dcd9b36d2390e57de

nicola-mazzucato-arm:
Jenkins: ok to test

tarek-arm (approved):

PR #813, opened 2023-07-12
Fix "all-log-level" defect

niaste-arm:
Fix defects raised by the 'scp-tests-ci-all-log-levels' Jenkins Job in product/n1sdp, product/tc2, and morello when CPPcheck raises unused variable warnings when log level parameter is set to either 'DISABLED' or 'CRIT'.


tarek-arm:
Jenkins: OK to test

PR #814, opened 2023-07-12
module/power_domain: Update all power state variables upon report

mohamedasaker-arm:
pd->current_state is updated upon receiving a transition report; other state variables are not. This can be valid when the report results from a request. However, this won't be valid for non-requested transitions such as wakeup. Hence, update all state variables accordingly.


nicola-mazzucato-arm:
Jenkins: ok to test

tarek-arm (approved):

PR #817, opened 2023-07-19
Fix Docker warnings and dependencies

ylpoon-arm:
- Fix warnings when running check scripts
- Fix project build for rcar in docker
- Install python and ruby dependencies on image build


nicola-mazzucato-arm:
Jenkins: ok to test

PR #821, opened 2023-07-27
assert: Add FWK_WEAK to __assert_func().

yaronm-hailo:
Allow overriding __assert_func for custom assert behaviour.

### DIFF
--- a/framework/src/assert.c
+++ b/framework/src/assert.c
@@ -6,6 +6,7 @@
  */
 
 #include <assert.h>
+#include <fwk_attributes.h>
 
 #if defined(__ARMCC_VERSION)
 void __aeabi_assert(const char *expr, const char *file, int line)
@@ -15,7 +16,7 @@ void __aeabi_assert(const char *expr, const char *file, int line)
     }
 }
 #elif defined(__NEWLIB__)
-void __assert_func(
+FWK_WEAK void __assert_func(
     const char *file,
     int line,
     const char *function,


tarek-arm:
Hi @yaronm-hailo,

Could you kindly share additional details on what prompted this change?

Thanks,
Tarek

yaronm-hailo:
Hi @tarek-arm,

Our team wanted the assert details (file, line, function name, assert string) to be logged to UART before going into the infinite loop to make debugging of asserts easier.
We preferred not to change the current default behaviour in the framework because we are not sure this is the desired behavior of other users of SCP-Firmware, so we opted for adding FWK_WEAK to the function __assert_func to allow overriding it.

Our implementation is the following:
```
static bool assert_already_happened = false;

void __assert_func(const char *file, int line, const char *function, const char *assertion)
{
    if (!assert_already_happened) {
        assert_already_happened = true;
        FWK_LOG_ERR("*ASSERTION OCCURED* on the following stage:");
        FWK_LOG_ERR("%s:%d:%s: (%s)", file, line, function, assertion);
    }
    fwk_interrupt_global_disable();
    while (1) {
        continue;
    }
}
```

tarek-arm:
Hi @yaronm-hailo, Thanks for the clarification. We are working on an internal solution to this problem. We will get back to you on this.

AhmedImbabyGadallah:
ok for me

AhmedImbabyGadallah (approved):

PR #823, opened 2023-07-27
system_power: allow having 0 elements

yaronm-hailo:
Remove fwk_check that the element_count > 0, which prevents a system without PPUs from using the system_power module.

### DIFF
--- a/module/system_power/src/mod_system_power.c
+++ b/module/system_power/src/mod_system_power.c
@@ -408,14 +408,15 @@ static int system_power_mod_init(fwk_id_t module_id,
     const struct mod_system_power_config *config;
 
     fwk_assert(data != NULL);
-    fwk_check(element_count > 0);
 
     system_power_ctx.config = config = data;
     system_power_ctx.mod_pd_system_id = FWK_ID_NONE;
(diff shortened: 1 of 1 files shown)


module/system_power/src/mod_system_power.c
```diff
@@ -408,7 +408,6 @@ static int system_power_mod_init(fwk_id_t module_id,
     const struct mod_system_power_config *config;
 
     fwk_assert(data != NULL);
-    fwk_check(element_count > 0);
```
leandro-arm, line 411: `element_count == 0` would leave to unexpected behavior on `fwk_mm_calloc` call on line 418.  And since the module is not expecting to have zero elements some parts of the code need to be reworked

leandro-arm (changes requested):

yaronm-hailo:
Hi @leandro-arm,
I added a check around the `fwk_mm_calloc` call.

I have reviewed the usages of `dev_count` and `dev_ctx_table`, and all are compatible with `dev_count` being 0 and `dev_ctx_table` is only accessed if there are elements.

tarek-arm (approved):

PR #846, opened 2023-09-25
module/power_domain: Only respond to `SYSTEM_SHUTDOWN` if necessary

leandro-arm:
Upon receiving a `PD_EVENT_IDX_SYSTEM_SHUTDOWN` event, `check_and_notify_system_shutdown()` checks for subscribers to the `MOD_PD_NOTIFICATION_IDX_PRE_SHUTDOWN` notification and notifies them. If there are subscribers, the `is_delayed_response` flag of the `PD_EVENT_IDX_SYSTEM_SHUTDOWN` event is set in order to wait for the subscribers answer. However, this message does not require an answer as the `response_requested` field of this event is never set.

Check that:
- The `response_requested` flag is set in the `event` input parameter in order to only answer if a response was requested.
- `mod_pd_ctx.system_shutdown.cookie` was set if a delayed answer is expected. The cookie and the `ongoing` are then reset.

### DIFF
--- a/module/power_domain/src/mod_power_domain.c
+++ b/module/power_domain/src/mod_power_domain.c
@@ -187,8 +187,17 @@ struct system_shutdown_ctx {
     /* Type of system shutdown */
     enum mod_pd_system_shutdown system_shutdown;
 
-    /* Cookie of the event to respond to */
+    /* Flag indicating if a response is requested or not */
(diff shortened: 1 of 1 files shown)


mohamedasaker-arm (approved):
Thanks for the change 🚀!

PR #848, opened 2023-09-26
ppu power status

khaledAhmed-arm:
After set power received, PPU report the current power state hardcoded without
any check the HW state.

Return the current power status if the set power return in successful
state otherwise get the current power status from direct reading HW.

Fix error messages logs by replacing [PD] by [PPU].


mohamedasaker-arm (approved):
Thanks for the change 🚀!

PR #850, opened 2023-09-28
Use of transport based fast channel interface.

girishpathak:
This PR requests groups two patch sets that enable the use of a new transport-based fast channel interface as documented in the module/transport/doc/transport.md(part of this commit). 

1. It introduces a new module fch_polled which implements a fast channel driver that polls(regular interval as configured by the platform) fast channel memory and directs callback to the appropriate protocol.
2. Re-factors existing scmi_perf code to use the new interface.


girishpathak:
Jenkins: ok to test

leandro-arm (approved):

PR #856, opened 2023-10-09
adding PID controller feature for Thermal Management

hisandeepsingh:
Adding Derivative term in PI Controller . 
Included the PID changes protected inside flag. Please review it and let me know if its looks good then we can do respective changes in test cases and in thermal doc thermal_mgmt.md

### DIFF
--- a/module/thermal_mgmt/src/mod_thermal_mgmt.c
+++ b/module/thermal_mgmt/src/mod_thermal_mgmt.c
@@ -34,6 +34,68 @@ struct mod_thermal_mgmt_actor_ctx *get_actor_ctx(
     return &dev_ctx->actor_ctx_table[actor];
 }
 
+
+#ifdef PID
+static void pid_control(fwk_id_t id)
+{
+    struct mod_thermal_mgmt_dev_ctx *dev_ctx;
+    int32_t err, terr, pid_power;
+    int32_t k_p, k_i,k_d;
+
+    dev_ctx = get_dev_ctx(id);
+
+    if (dev_ctx->cur_temp <
+        dev_ctx->config->pid_controller.switch_on_temperature) {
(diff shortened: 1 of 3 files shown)


mohamedasaker-arm:
jenkins: ok to test

module/thermal_mgmt/src/mod_thermal_mgmt.c
```diff
@@ -176,7 +239,11 @@ static int control_update(fwk_id_t id)
     if (dev_ctx->control_needs_update) {
         dev_ctx->control_needs_update = false;
         if (dev_ctx->config->thermal_actors_count > 0) {
-            pi_control(id);
+#ifdef PID
+            pid_control(id);
+#else
+           pi_control(id);
```
leandro-arm, line 245: Can you please run `clang-format` or check `tools/check_tabs.py` script?

leandro-arm (changes requested):
Can you also please update the top banner in all modified files to be
`Copyright <created year> - 2023`

module/thermal_mgmt/src/mod_thermal_mgmt.c
```diff
@@ -34,6 +34,68 @@ struct mod_thermal_mgmt_actor_ctx *get_actor_ctx(
     return &dev_ctx->actor_ctx_table[actor];
 }
 
+
+#ifdef PID
```
nicola-mazzucato-arm, line 38: IMO there's no need to create two implementations that are almost identical but the kd part.
I would suggest to add the derivative part in the pi_control as is, no additional #IFDEF.
Platforms that do not care about the derivative part can simply leave that coefficient to zero.

Then rename pi_controller() to pid_controller() please

PR #859, opened 2023-10-16
fwk_list: Add a watermark to monitor list max sizes

khaledAhmed-arm:
Enable queue monitors and tracing the maximum queue elements as a watermark for both isr and event queues.

To enable queue monitor and watermark for tracing, `SCP_ENABLE_FWK_EVENT_WATERMARK_TRACING` shall be true.


AhmedImbabyGadallah:
jenkins: ok to test.

PR #865, opened 2023-11-08
product/rdn2: IONCICLK ctrl/div and TCU/IONCI clock enable should be RW

lstarret:
IONCICLK_CTRL/IONCICLK_DIV1 were incorrectly marked as RO and WO, which breaks the read/modify/write behavior in mod_pik_clock.c.  Also updated TCUx_CLK_ENABLE and NCIx_CLK_ENABLE R/W attributes, even though not yet referened in the existing SW.

### DIFF
--- a/product/rdn2/include/system_pik.h
+++ b/product/rdn2/include/system_pik.h
@@ -1,6 +1,6 @@
 /*
  * Arm SCP/MCP Software
- * Copyright (c) 2020-2021, Arm Limited and Contributors. All rights reserved.
+ * Copyright (c) 2020-2023, Arm Limited and Contributors. All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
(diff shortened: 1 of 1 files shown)


thomas-abraham-arm (approved):
Thanks for the fix.

leandro-arm (changes requested):
@Joe-Zhucg Thanks for the patch, it looks good to me too, 

Can you also please update the top banner in all modified files to be
Copyright <created year> - 2023. ? to make it pass the CI
You can check it by running tools/check_copyright.py

Thanks!

leandro-arm:
Jenkins: ok to test

leandro-arm (approved):

nicola-mazzucato-arm:
s /referened /referenced - feel free to ignore

lstarret:
Whoops... fixed.

PR #872, opened 2023-11-14
Pre-requisite patches for extending RD-Fremont platform support

thomas-abraham-arm:
This pull request contains the set of patches that are pre-requisite patches for extending RD-Fremont platform support with updates for SCP, MCP and LCP.


girishpathak:
Jenkins: ok to test

PR #873, opened 2023-11-14
check_copyright: Exclude moved files from checks

nicola-mazzucato-arm:
Files that are moved from one directory into another with 100% similarity should be excluded from the copyright check, in other words, they should not need an updated copyright year.

Add a new git operation and introduce a further comparison for files that are simply moved across directories to be excluded from the copyright check in the loop.

Change-Id: Ia9f4ece2d467eee61fc6a6afd8a07ba8339d54c6


mohamedasaker-arm (approved):
Thanks for the change 🚀

PR #875, opened 2023-11-20
morello: print the temperature sensor values during boot

sah01Kaushal:
Currently, on SCP, there is no way for users to get CPU temperature values during or after boot. This change will show reference temperature values during boot, so that the alarm or shutdown threshold does not come as a surprise.

Change-Id: I6079a750d4aeaa7768784aee640919cf2faa76b0


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #876, opened 2023-11-20
 rdn2/pcie_integ_ctrl: Move module from common module directory to rdn2

nissha03:
PCIe Integration controller is only present in RD-N2 family of platforms and other platform does not use this code.
Move the module to the RD-N2 platform module directory.


nissha03:
> @nissha03 please rename this PR and provide a brief description, thanks

Done

mohamedasaker-arm:
jenkins: ok to test

PR #878, opened 2023-11-21
cmn700: fix address and size alignment check.

nissha03:
Fix the logical error in the cmn700 address alignment check. Due to the logical error, the check only failed if the address and size were unaligned.

Updated the logic to allow it to fail if either value is unaligned. Also fixed the wrong address mask calculation.


nicola-mazzucato-arm (approved):

nicola-mazzucato-arm:
Jenkins: ok to test

PR #881, opened 2023-11-28
Halt SCP boot if DDR4 training fails for morello and n1sdp

sah01Kaushal:
Currently, the SCP boot continues if the DDR4 training fails, the boot then later hangs when it reaches tf-a stage. These patches will halt the boot in SCP itself.


nicola-mazzucato-arm:
Jenkins: ok to test

nicola-mazzucato-arm (approved):

PR #884, opened 2023-11-29
Update MPMM to use AMU driver and add TC2 AMU mmap config

wenping-arm:
- Modify MPMM to use AMU driver API interface instead of directly accessing the memory to read AMU counters.
- Update TC2 MPMM configuration to include AMU driver API interface.
- Add TC2 AMU memory mapped driver configuration.


wenping-arm:
Fix/amu mmap allocation #883 has to be merged first before merging this PR.

PR #885, opened 2023-11-29
mod_transport: Make out-band message support optional

AndyJGraham:
The goal of this change is to make out-band message support option through use of conditional compilation.

The compilation flag that was added is:
BUILD_HAS_OUTBAND_MSG_SUPPORT

All platforms that use "mod_transport" have been
updated to have this new flag by default.

Change-Id: I903f77ada1d61128ee4b76cb130cfe4fc4ff81ab


nicola-mazzucato-arm:
Jenkins: ok to test

PR #888, opened 2023-12-08
fix/unit-test: add variable initialization

leandro-arm:
There are some variables unitilialized for testing resulting in unexpected behaviour.
This patch fixes that giving them a default `0` value to ensure that all bytes are zeroed.

Change-Id: I29265f415bd9f9b1462884da7bc97ef83f1581a0


mohamedasaker-arm (approved):
Thanks for the fix 🚀!

